Issue #542, opened 2014-07-29
Survey v1 UI: Reducing options


bbasso:
In the Advanced Options, Grid Options tab, Overshoot should just be one option, and the same on each side of the polygon.

Overlap and Sidelap can be the same number as well, as long as they are area-wise, not length-wise. ie. 60% overlap means 60% of the area is the same, regardless of if it's on the side or the top of the sensor footprint.

![image](https://cloud.githubusercontent.com/assets/4072292/3730120/0575588e-16ca-11e4-9dd4-7abff3bd5987.png)

Issue #506, opened 2014-07-03
[request] Survey (Grid) Jump two lines to get a bigger turning radius


Yups1337:
hey folks!

This is my second request to improve the Survey capabilities with MissionPlanner and Arduplane.

To get a good GroudSampleDistance you have to fly at low altitudes (~100m) and the MissionLines very close to each other, about 30-50m. With a regular plane you have a turning radius about 75m. If you want to get a good turn, the lines must have a distance of 150m. This is far to much!

![turningproblem](https://cloud.githubusercontent.com/assets/3913687/3475423/b895c13a-02f0-11e4-8eb1-dff7677c127e.jpg)
Screenshot of a testmission. You can see that the turning circle is much bigger than the line distance. The plane has big problems to fly on its flightpath!!

But you could jump over to the next or even more. 

![mp2](https://cloud.githubusercontent.com/assets/3913687/3475440/0f1ff818-02f1-11e4-951c-f1b1fa07bb0c.jpg)

This could be a solution. Not perfect but way better than the regular mission design. What do you think?

Edit:
ReMapHI has a better description for the flight pattern, it is called zamboni pattern. 
Here is a screenshot by SteveJoyce
![kmloverlay](https://cloud.githubusercontent.com/assets/3913687/3488695/d19968c2-04fc-11e4-9736-33559f21b276.PNG)

iskess:
It looks like the Zamboni pattern was finally added by Pull Request #777

PR #777, opened 2015-01-11
Survey Grid generation improvement. Alternating Lanes

wjax:
- Improve Grid generation by calculating closestLines to point by a
perpendicular proyection
- Add basic new feature to allow user to Alternate lanes. Useful for
plane users as turn radius is increased. No need to overshoot so much

### DIFF
--- a/ExtLibs/Grid/Grid.cs
+++ b/ExtLibs/Grid/Grid.cs
@@ -70,7 +70,7 @@ namespace MissionPlanner
             //map.Invalidate();
         }
 
-        public static List<PointLatLngAlt> CreateGrid(List<PointLatLngAlt> polygon, double altitude, double distance, double spacing, double angle, double overshoot1,double overshoot2, StartPosition startpos, bool shutter)
+        public static List<PointLatLngAlt> CreateGrid(List<PointLatLngAlt> polygon, double altitude, double distance, double spacing, double angle, double overshoot1,double overshoot2, StartPosition startpos, bool shutter, float minLaneSeparation)
         {
             if (spacing < 10 && spacing != 0)
                 spacing = 10;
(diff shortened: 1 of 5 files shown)


wjax:
First of more to come additions to improve grid generation for planes.
Following would allow a maximum turning angle and thus calculating lane separation and waypoints order.

iskess:
This was originally requested in #506 and the solution identified as a Zamboni pattern by @RemapHi.

wjax:
Hi iskess

My solution does not use zamboni pattern. I did not know about it until you just pointed me there.
The solution in this pull request allows the user to select the minimum separation (in number of lanes) to be used by Mission Planner to generate the grid to fly.

I have tested it and it works quite well.
I was planning to add a "minimum radius" to the grid generation algorithm so the grid and waypoints dont force an aircraft to do something it cannot acomplish.

wjax:
Take a look

https://drive.google.com/file/d/0BzRzbaLTcOpbeVBwUVFPMktWSVE/view?usp=sharing

iskess:
Jesus,
I like the idea of a minimum radius. Perhaps it could be taken a step further and the user could enter a projected wind velocity. The algorithm will look at the tailwind vector component on the downwind turn, use the cruise speed, use the max bank angle parameter, and calculate the minimum turn radius. Using wind has other advantages, such as identifying the maximum triggering speed to see if we will exceed the camera's capabilities on the downwind leg. There is a good discussion on issue #542 .
As mentioned by @bbasso, it would be very interesting and useful to compare the ground track length of three methods; Yours, Zamboni's, and the current overshoot method. 
Thank you for working on this. There hasn't been any Survey Grid development since @tristannovak  made his great contributions. 

Iam

PR #794, opened 2015-02-06
*: forms "Display this"

kwist-sgr:
1. Dialogues "Display this" are shown modally.
2. Sorting options in alphabetical order
3. Fix clipped text in the names of the parameters
4. Fix setting the initial checked state for QuickView
5. Fix setting the initial checked checkboxes colour state

### DIFF
--- a/GCSViews/FlightData.cs
+++ b/GCSViews/FlightData.cs
@@ -512,7 +512,9 @@ namespace MissionPlanner.GCSViews
                         QuickView QV = (QuickView)ctls[0];
 
                         // set description and unit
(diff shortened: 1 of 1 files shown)


meee1:
what is that change intending to do?

kwist-sgr:
1. Dialogues "Display this" are shown modally.
2. Sorting options in alphabetical order
3. Fix clipped text in the names of the parameters
4. Fix setting the initial checked state for QuickView
5. Fix setting the initial checked checkboxes colour state

meee1:
done

PR #795, opened 2015-02-06
Localization: update zh-Hans translation

SkiTiSu:
Wizard fully translated.
Add UTF-8 support for config.


SkiTiSu:
Here is a question, three survey(s)' Form can't support Localization. I added the .zh-Hans.resx for them but they could't become zh-Hans when I use them (the CurrentUICulture of them are "zh-CN"). Would you do me a favor? Thanks a lot.
By the way, today is Chinese New Year's Eve. Warm hearted wishes for a happy New Year filled with all your favorite things.

PR #836, opened 2015-04-08
Adding log analyser

bionicbone:


bionicbone:
Sorry Michael, that was not suppose to do that against your master :(

PR #843, opened 2015-04-16
Uploader refactor: broke up enum values into two enums

magicrub:
Uploader: broke up enum values into two enums (created ne Info enum) and moved others to consts due to improper enum use in "Code".

non-function code change


meee1:
merged

magicrub:
Thanks!

PR #844, opened 2015-04-16
Log: DoubleClick log item to apply user-defined offset and scalar

magicrub:
When plotting data it is sometimes nice to add a scalar and/or an offset. Good examples are when comparing GPS alt vs baro alt vs lidar alt.
Any DataFlash .bin/.log logged value can have an arbitrary scalar and/or offset applied when plotted.
Enter modifier then value, they are applied in the order you provide. Modifiers are x + - /
Example: Convert cm to to m with an offset of 50: '/100 +50' or 'x0.01 +50' or '*0.01,+50'

### DIFF
--- a/Log/LogBrowse.cs
+++ b/Log/LogBrowse.cs
@@ -44,6 +44,104 @@ namespace MissionPlanner.Log
         GMapOverlay mapoverlay;
 		GMapOverlay markeroverlay;
 		LineObj m_cursorLine = null;
+        Hashtable dataModifierHash = new Hashtable();
+
+        class DataModifer
+        {
+            private bool isValid;
+            public string commandString;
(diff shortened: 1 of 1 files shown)


magicrub:
Bump

magicrub:
Michael, Do you have any questions about this that I can answer for you?

meee1:
can I request we use a null check instead of hardcoding default template
ie remove
 dataModifier = new DataModifer("x1+0");

this will improve performance than the current method.

magicrub:
Sure. I can do the edit tomorrow or the day after unless you beat me to it.

magicrub:
Done!

meee1:
merged

PR #888, opened 2015-06-01
Fix Rel vs AMSL Altitude Logic

jmachuca77:
The use of relative altitude vs AMSL was reversed in relation to the
check box. If the “Use AMSL” check box is checked it should write AMSL
to the pictures.


meee1:
merged thanks

PR #898, opened 2015-06-10
Added buttons and logic for binding of reciever

EShamaev:
ConfigRadioInput has three buttons now for binding the receiver.
Support in ardupilot was implemented (Dsmx bind #2394)
https://github.com/diydrones/ardupilot/pull/2394


EShamaev:
The build failed because of Mono JIT compiler problem at Travis side.

tridge:
applied, thanks!
(with a few formatting and error checking changes)

EShamaev:
I could not find the proposed features in Mission Planner's master branch. Is there something I can do with the code to make it right for the project?

limhyon:
This feature is not working on Pixracer. Can anyone verify my test? see if you have same problem or not.

PR #931, opened 2015-07-13
Keyboard Controller

denisdifazio:
Hello,

I would like to introduce to Mission Planner my Keyboard Controller tool. It allows the user to pilot drones using just the keyboard keys. It works pretty much the same way Joystick's tool does.
Furthermore, it is possible to close or hide its windows without interrupting its functionality.

Here is a image of the form which is added to Mission Planner:
form


meee1:
I'm hesitate to merge this, as there is no safety around it. and could see it being misunderstood, and causing an accident.

denisdifazio:
Hi Michael, I can work on the safety issues. Could you provide me information on what should be improved?

mzm0018:
I think it's a good idea, but we need to consider the signal delay between the autopilot and the computer through the radio.

gediminasgu:
I think that could be implemented as a separate Mission Planner plugin but not the part of the application.

jace26:
this would be super useful for simulator runs in particular.

RisingPhoenix247:
Does this work?? @denisdifazio

PR #968, opened 2015-08-14
Add confirmation to Help->Check Updates and fixed issues with Flight Planner tile prefetch

Hangarspace:
The 'Help->Check Updates' applies any pending update without providing the user with the ability to decline the update. Any potentially destructive action should be preceded with a confirmation.

The action now calls the 'Update.CheckForUpdate() method which already has the 'prompt-on-update-found' dialog. This method was also updated to  take an optional boolean parameter 'NotifyNoUpdate' that will optionally notify the user if no update is found. Without this, then there would be no feedback to the user that anything is or has happened.


meee1:
merged thanks

PR #993, opened 2015-09-09
Test and stop all motors buttons

marcoarruda:
Two more buttons at Motor test view. One to test all motors at the same time and another to stop all motors.


meee1:
merged thanks

PR #1002, opened 2015-09-14
Add "Abort Alt" description to NAV_LAND mission item

magicrub:
Add "Abort Alt" description to NAV_LAND mission item p1.


meee1:
done.

PR #1012, opened 2015-09-22
Fixed GeoRef by time offset

jmachuca77:
Fixed typo on the MAVlink message offset detection for Relative
Altitude that was causing the offset calculation to fail.
Added a check to ignore GPS messages that have the default time so that
the offset calculation average is correct.

### DIFF
--- a/GeoRef/georefimage.cs
+++ b/GeoRef/georefimage.cs
@@ -247,7 +247,7 @@ namespace MissionPlanner.GeoRef
                             int latindex = dflog.FindMessageOffset("GPS", "Lat");
                             int lngindex = dflog.FindMessageOffset("GPS", "Lng");
                             int altindex = dflog.FindMessageOffset("GPS", "Alt");
-                            int raltindex = dflog.FindMessageOffset("GPS", "RAlt");
+                            int raltindex = dflog.FindMessageOffset("GPS", "RelAlt");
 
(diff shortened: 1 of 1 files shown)


GeoRef/georefimage.cs
```diff
@@ -247,7 +247,7 @@ private DateTime getPhotoTime(string fn)
                             int latindex = dflog.FindMessageOffset("GPS", "Lat");
                             int lngindex = dflog.FindMessageOffset("GPS", "Lng");
                             int altindex = dflog.FindMessageOffset("GPS", "Alt");
-                            int raltindex = dflog.FindMessageOffset("GPS", "RAlt");
+                            int raltindex = dflog.FindMessageOffset("GPS", "RelAlt");
```
meee1: where did this change come from?
https://github.com/diydrones/ardupilot/blob/master/libraries/DataFlash/DataFlash.h#L723

meee1:
ive added here with changes
8fd8f1e6615de792c2e1b80afa7dbf09245de9da

GeoRef/georefimage.cs
jmachuca77: Interesting, its different on Solo:
https://github.com/3drobotics/ardupilot-solo/blob/master/libraries/DataFlash/DataFlash.h#L592

PR #1018, opened 2015-10-01
Some small fix and improvements

ZeSly:
Hello,

Here is the improvement I made :

1) When double-clicking on the graph, it shows the position on the map. I completly rewrite the function GetGPSFromRow() in Log/LogBrowser.cs

2) I add Min, Max and Mean value of the displayed curves. It is updated when you zoom on the graph

screenshot_logbrowser

3) I reviewed the tab order on the Extended Tuning screen. When tuning PID with a small laptop in the field, I thing it is easier to switch from a field to another with the tab key then with a mouse or a touchpad.


meee1:
in its current state, I cannot merge this, as I can see regression on some of the code.

meee1:
ive pushed what I can, but not all of it due to the regression.

ZeSly:
Can you explain what are these regression ? Thanks

meee1:
the removal of Msi/licence.rtf ?
and a few other comments inline in the commits.

I would suggest you look into how to use rebase, vs merge. it makes the commit tree a lot cleaner.

ZeSly:
Hi,

In my last commit (e5419f3a793e8bdcc453256eb213ada490ac2cb7) I reintegrated all the POS message processing. I will add searching for it in the GetGPSFromRow function as you suggest. If you compare Log/LogBrowse.cs from my repository with yours, you will see changes only in GetGPSFromRow and a small fix in GoToSample.

I did not understand what the merge have done with Msi/licence.rtf. At first, I thought it was removed from the diydrones repository. I did not edit it nor remove it.

meee1:
I cant use that though, as its a recommit of my own changes. hence why rebase is a better option.

ZeSly:
I added POS messages in GetGPSFromRow

meee1:
ive merged this, but had to merge some commits.
1121493593d36976c550cfe4bbf747c4217daf1b

meee1:
could I request you create a new branch for your latest changes?

PR #1021, opened 2015-10-04
New compass setup page

jacobsenmd:
This is a completely redesigned setup page for the compass. It updates the user interface to account for the fact that ArduPilot now supports up to three compasses. The new interface gives a much clearer presentation of how each compass is configured, and is adaptive to versions of firmware that support 1 vs 3 compasses. It also includes color-coded feedback on compass offsets for each compass. This also fixes a bug where newly calibrated compass offsets were not refreshed in the parameter list. Finally, it updates the live compass calibration form to show errors for 1 or 3 compasses depending on what's supported by the firmware. I believe all these changes will add a great deal of clarity to compass configuration.

This is my first GitHub pull request. I did not squish commits in case anyone wants to review the sequence of work, and I am not sure the standard on this. Please let me know if there is anything I can/should do differently.

Some outstanding issues involved in this commit:
- I am not clear how localization works in Mission Planner, so it is possible that the new interface will break some localization. I'm guessing this isn't a huge issue, based on how many strings were hard-coded already
- Compass offsets turn yellow if any dimension exceeds 400 and red if any dimension exceeds 600. The 600 value was based on input from Randy about when the compass indicates an error. The 400 was arbitrary. I'm open to changing these thresholds.
- I do not understand why the previous code configures the Pixhawk orientation the way it does. I kept this configuration code as is, but do not understand it and think it deserves a review. In particular, I don't understand why the orientation is different for different versions of firmware, and why the code for the old firmware specifies a different orientation but then sets COMPASS_EXTERNAL to 0... which would cause the Pixhawk to ignore the specified orientation.


meee1:
commit here
https://github.com/diydrones/MissionPlanner/commit/575cf82eb076d4b067eaa2f5c27ed63b90ce2527

PR #1023, opened 2015-10-06
New georef function

Chatmetaleux:
As I have bad logs from flights (CAM msgs > photos), for me the offset method is not enough to solve my synchronization problem so I wrote a new function based on the existing methods found in georefimage.cs. The new function here is called by the CAM sync button (if numbers of CAM/picture mismatch).

I don't know if I should already pull this code...Michael ?

Btw, i raised the limit of 4 CAM/pictures to 10 to find the offset with the "Estimate offset" button.

-----
Mission Planner: Photos and CAM synchronization (documentation)

The first trivial case is when the number of CAM msg from the log and the number of photos match.
We speak here about the more complex way to synchronize photos and CAM when the number of CAM msg is bigger than the number of photos. 
First of all, we need to take in mind some facts:
-	The photo time accuracy is at the second.
-	CAM and GPS msg accuracy are at 200 millisec.
-	Time system of GPS and camera are not synchronized.
-	The shutter lag is variable (if no manual shutter lag is set -> “auto focus mode”)
-	In MP, we can use an offset(=Photo time – CAM time) to apply to CAM or GPS msg. MP can compute an offset with the 4 first CAM/photos (we assume here that there is no photo missed in this four, if not the offset would rise). But this way is not efficient, it needs to adapt the offset manually.

The problem is as follow:
	
mpalgo

•	T0p and T0c are different.
•	m >= n.
•	ΔT1 is the first offset. We assume that’s correct (no missing photo for the first CAM) but not the best offset; 
•	If not correct, all photos locations would be shifted. We could estimate an offset based on first 5 CAM msgs and the first photo, but (after tried that) there are some cases where the shifted offset matches better than the real one. It’s because of the repeat sequence of cam+photos in time (every 200 millisec and seconds). So having a good first start is primordial.
•	ΔT1->m is variable.
•	Some photos are taken at the same time, so we can’t determine 2 different locations, it will be overlapping.
•	Some photos are missing (not taken, or CAM msg too close).

So we need an algorithm that can match photo to the good CAM msg and compute a best offset to match all photos:
From the first offset ΔT1, we compute the matching CAM/photo time between [ΔT1 - 1 sec ; ΔT1  + 1 sec] – called Offset0->end - with a step of 200 ms (CAM/GPS accuracy).
The value of 1 second is a “big maximum” value of the shutter lag.
For all this Offset0->end we compute a score value, to minimize the average ΔT1->m .
To take more into account the big ΔTi (avoid them), we count the ΔT^3 average as a score.
To match a photo and a CAM, we look for a CAM msg 2 seconds around the photo time (corrected with an Offseti). Note that the value of “2 seconds” could be down to 1 (maximum shutter lag and photo time accuracy at 1 sec); If there is no CAM found, we don’t take this photo into account.
Note : as photo time accuracy is lower to CAM/GPS accuracy, that means that a good CAM msg could be AFTER a photo time, so we can’t use other method to match the CAM and photo time.
Finally, we chose the Offseti that have the most matched photos and the lowest score; and then we apply it to the CAM/photo detection like MP does with the option “Use CAM msg + Time offset way”
Another change is the way to find a CAM/GPS msg from a corrected time (offset), the old method takes the first CAM/GPS from the corrected time within a limit of X sec after. Mine looks for both directions from the corrected time.
Using AMSL Altitude True
Reading log for GPS Messages in order to get AMSL Altitude
Log Read for GPS Messages
Reading log for CAM Messages
Log Read with - 905 - CAM Messages found
Read images
Images read : 782
CAM Msgs and Files discrepancy. Sam's method instead !
Reading log for CAM Messages
Read images
Images read : 782
Storing photos time.....
offset 7.4 ; Score :-0.197897520661156 ; NbrP : 605
offset 7.6 ; Score :-0.0862478777589136 ; NbrP : 589
offset 7.8 ; Score :0.0417564322469983 ; NbrP : 583
offset 8 ; Score :0.110278637770897 ; NbrP : 646
offset 8.2 ; Score :0.114098550724637 ; NbrP : 690
offset 8.4 ; Score :0.134947368421052 ; NbrP : 760
offset 8.6 ; Score :0.0671918158567777 ; NbrP : 782
offset 8.8 ; Score :0.00701790281329924 ; NbrP : 782
offset 9 ; Score :-0.0497186700767264 ; NbrP : 782
offset 9.2 ; Score :-0.151017902813298 ; NbrP : 782
offset 9.4 ; Score :-0.153789321789321 ; NbrP : 693
Best offset :8.8....
Photo DSC05657 PROCESSED with CAM position found 400 ms away
Photo DSC05658 PROCESSED with CAM position found 0 ms away
Photo DSC05659 PROCESSED with CAM position found -400 ms away
Photo DSC05660 PROCESSED with CAM position found 200 ms away
Photo DSC05661 PROCESSED with CAM position found 0 ms away
Photo DSC05662 PROCESSED with CAM position found -200 ms away
Photo DSC05663 PROCESSED with CAM position found -400 ms away
Photo DSC05664 PROCESSED with CAM position found 400 ms away
Photo DSC05665 PROCESSED with CAM position found 200 ms away
...

This algorithm has been tested on 20 flights (Falcon), and checked manually with the first and last photos (take out and landing) + flight line
Note that the accuracy of the GPS and this result is still relative and we should use the result of alignment with Photoscan and ground marker with better accuracy to really use the location and orientation of photos.


Chatmetaleux:
I need to check and adapt it for a better algorithm...

PR #1047, opened 2015-10-29
RadioCalibration: Do not write values if no change occurs

DroneBuster:
This change makes radio calibration to not write new values if difference between min and max is less than 10 (could be changed). This saves from struggle for example then you have something set on your channel function to something like flapperons. Always after calibration you have to go back and reenter values.

Also this lets to re-calibrate 1 channel only.


meee1:
I don't really like this change, as it can create out of range values. and doesn't really make any operation difference.

PR #1050, opened 2015-10-30
Make MissionPlanner waypoint lists compatible with QGroundControl

mairsbw:
First commit improves compatibility while not altering MissionPlanner functionality at all. Second commit changes the functionality where users would need to use .waypoints so that transitioning between the two programs is then seamless.

### DIFF
--- a/GCSViews/FlightPlanner.cs
+++ b/GCSViews/FlightPlanner.cs
@@ -1509,7 +1509,7 @@ namespace MissionPlanner.GCSViews
         {
             using (SaveFileDialog fd = new SaveFileDialog())
             {
-                fd.Filter = "Ardupilot Mission (*.txt)|*.*";
+                fd.Filter = "Ardupilot Mission (*.txt)|(*.waypoints)|*.*";
                 fd.DefaultExt = ".txt";
                 fd.FileName = wpfilename;
                 DialogResult result = fd.ShowDialog();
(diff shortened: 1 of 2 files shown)


meee1:
done

billbonney:
Doesn't this change just make all tools and everything that reads 110 files incompatible with Mission Planner saved files (no changes in the file format have happened, so it seem this fix is in the wrong code) i.e. QGC should be modified to load a QGC 110 marked file?

mairsbw:
Well, it could be that Mission Planner has been using a 1.2 version of the mission file this whole time and wasn't broadcasting it correctly. Unfortunately this file format is not well specified, and actually both Mission Planner and QGroundControl are not in agreement over many of the finer details. Do you know of any tools that will choke on a 1.2 file version?

I'll write up a spec and send it out to the qgroundcontrol mailing list and see if we can get this standardized and then we can all implement that standard.

mairsbw:
They've attempted to document it [here](http://qgroundcontrol.org/mavlink/waypoint_protocol#waypoint_file_format), but it's actually incorrect.

mairsbw:
Additionally, I believe that MissionPlanner saves the home location as a first "mission" in the file, whereas QGC does not. (see https://github.com/mairsbw/MissionPlanner/blob/266b8230f8f10d0add65e991643ef86f71271b31/GCSViews/FlightPlanner.cs#L1525)

billbonney:
APM uses the first location in the mission list as Home location. This is true. But not relevant for this discussion
AndroPilot, Tower (Droid Planner), uGCS, APM Planner 2.0 (though in that case we just check for 110>) and anybody else reading those files are potentially impacted. This is for a change of a version number from 110 to 120. Noting nothing else as changed.

It easy enough to talk with the QGC guys, or submit a PR for them to accept QGC 110 files into their waypoint manager. Which results in 1 change, and does not to impact the whole ecosystem. 

If you have new mission format you want to push lets do that, but intentionally breaking whats already out there without a good reason doesn't seem like a good idea.

billbonney:
PS: I can help you a PR for QGC to remain compatible.

PR #1052, opened 2015-11-05
Resharper run to reformat code

kjeremy:
In looking through the code I was having some trouble with the existing indentation in VS2013 so I ran Resharper on some of the libraries.  There are no code changes, just formatting.


meee1:
no longer needed

PR #1060, opened 2015-11-19
WMS Server URL with query strings and Input Box UI

haripradhan:
Added functionality to accept WMS server URL with query strings, Displayed WMS layer title instead of the layer name, Auto-sized an input box displaying the list of WMS layers vertically.

verticalautosize


meee1:
merged thanks

PR #1062, opened 2015-11-23
Make waypoints appear on flight data screen on WP read

DroneBuster:
I ran into problem that waypoints aren't displayed on Flight Data screen when they were read from MAV. They appear only when they are written to MAV. I don't know if this is intentional, but this is quite problem for me when i'm using large missions (~700wp's).


meee1:
Hi, thanks for the pull request, I ended up using my own code for this one
code is here
3b24c639

PR #1065, opened 2015-11-23
Allow a plugin to insert waypoints anywhere in the list

kjeremy:
Introduces `Host.InsertWP` which allows the caller to insert a waypoint anywhere in the list.


meee1:
merged thanks

PR #1066, opened 2015-11-24
GMapMarker: Make Position virtual

kjeremy:
This allows custom positioning logic per marker: for instance a marker type could have it's position set once and not be allowed to move.


meee1:
merged thanks

PR #1078, opened 2015-12-11
VRBRAIN USB drivers update

LukeMike:
Added detection for firmware management of new VR Brain Core 1.0 and deleted unnecessary VR boards.
Updated description of some USB Drivers and added missing drivers


meee1:
Commited Thanks

PR #1097, opened 2016-01-05
Initialize theme colors to match the default theme 'BurntKermit'

kjeremy:
Gives the user something useful in the WPF designer.


meee1:
merged thanks

PR #1101, opened 2016-01-11
Make URLs clickable in param description

magicrub:
closes https://github.com/diydrones/MissionPlanner/issues/1100


meee1:
merged thanks

PR #1102, opened 2016-01-11
FlightData: Add the ability to swap the map and HUD

kjeremy:
Allow the user to right click on the HUD and swap positions with the map.


meee1:
merged thanks

PR #1105, opened 2016-01-13
Video resolution fix

kjeremy:
Fixes video quality issues when using VLC to stream hi-def video and improves HUD rendering performance.


meee1:
does this code deal with the hud being resized? this is why it was a locked size originally, as i hadn't fixed that yet.

kjeremy:
Yes the HUD resizes just fine with this.  It doesn't alias anymore with this change.  Non power of two textures have been supported since OpenGL 2.0 (2004).  If you want I can add a check and fallback to the resize if the hardware doesn't support it.

kjeremy:
I've added a check for non-power of two texture support that will fall back to the old behavior if unsupported

meee1:
merged thanks

PR #1106, opened 2016-01-14
Disable joystick functionality when joystick is not plugged in

mwieliczki:
This fixes the problem where the joystick can be enabled when it is not connected.  Previously ardupilot would receive joystick messages when the joystick was not plugged in.


meee1:
merged thanks

mwieliczki:
no problem

PR #1107, opened 2016-01-19
Reduce build times

kjeremy:
Reduce build times by only copying files if newer.  Previously when building `MissionPlanner` both `MetadataExtractor` and `px4uploader` would **always** build.  The added build time is especially noticeable when building plugins since they depend on `MissionPlanner`.  By only copying rarely updated resources if they are newer instead of always we prevent the compilation of `MetadataExtractor` and `px4uploader` unless files have changed.


meee1:
commited thanks

PR #1108, opened 2016-01-19
Reduce memory allocation and speed up HUD drawing

kjeremy:
Moves pen/brush allocations to instance variables and disposes of temporary
brushes when used.  This reduces memory peak allocation by 50MB while
streaming in the HUD and keeps usage more stable (fewer peaks and troughs).


meee1:
commited thanks

PR #1110, opened 2016-01-20
Fix hud artifacts

kjeremy:
Fixes the grey box and diagonal line through video on the hud.  Also cuts down on texture generation overhead by reusing the texture in DrawImage.


meee1:
commited thanks

PR #1114, opened 2016-01-25
Fix the build

kjeremy:
Fix the build with respect to 3DRadio -> Sikradio


meee1:
I already have changes pending for this, just forgot to push

PR #1116, opened 2016-01-27
Add angle (degrees) column in flight planner mission chart

magicrub:
Gradient is currently available but having angle there too is nice. Who uses gradients?!?!


iskess:
+1. As a pilot I always wished we had a real glide slope angle. However I do see how a % gradient would be more intuitive for non-pilots, and I've grown used to it. 
I think we should have both.

magicrub:
This adds the degrees, gradient is still there.

meee1:
added, thanks

PR #1117, opened 2016-01-27
Added "Abort Land" click box on main Flight Data Actions screen

magicrub:
Want to abort your landing without a transmitter? Just press this button and it will come around and try that landing again ALL BY ITSELF! Weeeee

Only works for Plane, tooltip says so.


meee1:
added, thanks

PR #1120, opened 2016-01-28
Use GMap.Net from nuget

kjeremy:
This should be easier to maintain and it is a lot faster in debug builds.


meee1:
there are custom changes. so don't think this can happen.

PR #1142, opened 2016-02-16
mavlink: generate mavlink on every build - do not store output

magicrub:
- this is how ArduPilot does it and it's better. Only store and edit the xml file, the rest becomes magic
- the csharp files are generated and should not be tracked by git, only the xml


magicrub:
This adds a Python project build dependency. I've added that to the wiki.

meee1:
I made some mods and merged

PR #1144, opened 2016-02-17
Add progress feedback during firmware upload

clovett:
Add progress feedback during firmware upload.
Also added missing DirectInput dll's since the DirectX redist link is broken on MSDN.


meee1:
Ive always tried to manage the app as a portable app, so moving the config file location I don't think is a good idea.
also the directx file I think would be better to update the link/prereqs, than include in git
https://www.microsoft.com/en-us/download/details.aspx?id=8109

clovett:
Good point, I will remove the directx dlls.

PR #1149, opened 2016-02-21
Cleanup Settings and add progress for firmware upload.

clovett:
This change adds progress feedback during firmware upload and attempts to cleanup the app level settings code a bit by adding new Settings class to wrap the config hashtable.


meee1:
Ive merged your changes, but have left file locations as is, for the moment.

PR #1153, opened 2016-02-23
Add AutoLifeguards in FlightPlanner

bboypupu:


meee1:
i think this would be better done as a MP plugin.

Issue #1162, opened 2016-03-01
Latest fails to build with Visual Studio


kjeremy:
I'm seeing a lot of build failures on `master` today with Visual Studio.

```
8>------ Build started: Project: SharpKml, Configuration: Debug Any CPU ------
3>C:\Program Files (x86)\MSBuild\14.0\bin\Microsoft.Common.CurrentVersion.targets(1249,5): error MSB3073: The command "rd /s /q "C:\Projects\MissionPlanner\ExtLibs\Mavlink\generator\Csharp"
3>C:\Program Files (x86)\MSBuild\14.0\bin\Microsoft.Common.CurrentVersion.targets(1249,5): error MSB3073: python C:\Projects\MissionPlanner\ExtLibs\Mavlink\generator\mavgen.py --lang=csharp -o "C:\Projects\MissionPlanner\ExtLibs\Mavlink\generator\Csharp" --wire-protocol=1.0 "C:\Projects\MissionPlanner\ExtLibs\Mavlink\generator\message_definitions\ardupilotmega.xml"  
3>C:\Program Files (x86)\MSBuild\14.0\bin\Microsoft.Common.CurrentVersion.targets(1249,5): error MSB3073: copy /y "C:\Projects\MissionPlanner\ExtLibs\Mavlink\generator\Csharp\ardupilotmega\mavlink.cs" "C:\Projects\MissionPlanner\ExtLibs\Mavlink\Mavlink.cs"
3>C:\Program Files (x86)\MSBuild\14.0\bin\Microsoft.Common.CurrentVersion.targets(1249,5): error MSB3073: " exited with code 1.
```
...
```
28>------ Build started: Project: ExtGuided, Configuration: Debug Any CPU ------
24>CSC : error CS0006: Metadata file 'C:\Projects\MissionPlanner\ExtLibs\Mavlink\bin\Debug\MAVLink.dll' could not be found
24>CSC : error CS0006: Metadata file 'C:\Projects\MissionPlanner\ExtLibs\Comms\bin\Debug\MissionPlanner.Comms.dll' could not be found
24>CSC : error CS0006: Metadata file 'C:\Projects\MissionPlanner\ExtLibs\Controls\bin\Debug\MissionPlanner.Controls.dll' could not be found
24>CSC : error CS0006: Metadata file 'C:\Projects\MissionPlanner\bin\Debug\MissionPlanner.exe' could not be found
[...]

PR #1163, opened 2016-03-02
Fix the build by adding missing parenthesis around print.

kjeremy:
Building with python 3.5.1 on Windows 7 (installed via chocolatey) errors out without them.

### DIFF
--- a/ExtLibs/Mavlink/generator/mavgen_csharp.py
+++ b/ExtLibs/Mavlink/generator/mavgen_csharp.py
@@ -85,7 +85,7 @@ def generate_message_header(f, xml):
             firstchar = re.search('^([0-9])', fe.name )
             if firstchar != None and firstchar.group():
                 fe.name = '_%s' % fe.name
-                print fe.name
+                print(fe.name)
            
     t.write(f, '''
 using System;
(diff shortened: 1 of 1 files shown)


kjeremy:
Fixes #1162

magicrub:
shouldn't this be fixed in upstream then pulled from there?

kjeremy:
I don't know how @meee1 wants to handle this sort of things.  Some of the ExtLibs stuff are actually modified for this project (like the maps).  Ideally the mavlink project would just be a nuget package or the generated output would be pulled into `MissionPlanner` so we wouldn't have to generate the csharp every time.

kjeremy:
@magicrub Looking at upstream it's completely different then what we have in `MissionPlanner`.  They also already use `print` with parenthesis

magicrub:
is it just newer?

kjeremy:
Compare what's in MissionPlanner with https://github.com/mavlink/mavlink/tree/master/pymavlink/generator (unless I'm looking at the wrong code which is entirely possible).

meee1:
MP uses a custom parser i wrote. and yes it has aged.

kjeremy:
@meee1 can this be merged?

meee1:
what version of python you using? under 2.7 it works fine.
I will merge just wondering whats changed.

meee1:
merged

kjeremy:
Thanks.  It was Python 3.5.1 which is the default provided by [chocolatey](https://chocolatey.org).

meee1:
merged, thanks

PR #1168, opened 2016-03-07
Log: reject large spikes in log files

magicrub:
when data has large spikes (1E30) it can make it hard to view certain values. This eliminates those spikes but only if you are processing your data with an offset/scaler. So, to remove then just use " *1 ". This is a hack. @meee1 do you have a better way to show the original data but then also be able to remove glitches without adding work to existing normal logs?


meee1:
merged. thanks

PR #1192, opened 2016-03-16
Use Assembly.LoadFrom when loading plugins

kjeremy:
Previously we would fail loading plugins if they were dependent on another DLL
in the directory (nuget packages, third-party libraries etc.).  Using LoadFrom
successfully loads those dependent assemblies.

### DIFF
--- a/Plugin/PluginLoader.cs
+++ b/Plugin/PluginLoader.cs
@@ -24,7 +24,7 @@ namespace MissionPlanner.Plugin
 
             try
             {
-                asm = Assembly.LoadFile(file);
+                asm = Assembly.LoadFrom(file);
             }
             catch (Exception)
             {


kjeremy:
This also fixes an issue where plugin translations weren't loaded.  I really don't know WHY this fixes it but I have a plugin with a korean translation and with this change it properly loads those resources.

meee1:
merged. thanks

meee1:
looks like I might revert this commit, as it breaks other plugins now

kjeremy:
That's no good. I tested with all of them built and they worked. What's
broken? Without this change I don't get translations in my plugin.

meee1:
the grid plugin/other no long work
ive made a mod here
https://github.com/diydrones/MissionPlanner/commit/8f3bceb79ce608a9ad541ad077202ded1a9d9cae

meee1:
and
aee6afe9

kjeremy:
@meee1  How does this break other plugins?  I don't notice any problems with this change.

PR #1196, opened 2016-03-19
Fix some bugs in the LogDownloadMavLink form

clovett:
Not sure why diff is showing more, it has something to do with newlines, hopefully you can just ignore those changes.  Are you using the git config autocrlf=true ?  Anyway I cleaned up some UI code to do with downloading logs from dataflash, fixing bugs in log output (newlines were messed up when you resize the form).  Fixing Form Layout using TableLayoutPanel so things size properly with different system font sizes.  Moving strings to resx so they can be localized.  Adding CancellationTokens to make background work cancellable. This last one needs to be done across the code base, still way to many ways you can make the Mission Planner UI hang... but this is a start.


magicrub:
You have a merge in your history, you must remove that. It will get rid of
all the other changes. Please also rebase on master.

clovett:
Ok, I will do that.

PR #1204, opened 2016-03-24
Added CameraJoyStick

pristinesource:
Added a CameraJoystick class so enable a second joystick to be added to
control camera operations (MPO).  This MPO control has safety features
that are configurable to only allow it to override the main controller
(AVO) if it hits a certain threshold.  If that threshold is set to 0,
then the MPO has 100% control and the AVO has 0% over that channel.


pristinesource:
I added a second commit to fix the cameraJoyStick writing to the override channels and I also added a rate conversion option for tilt and zoom.

pristinesource:
Please wait to merge this, I found another issue and I am working on a fix for it.  I will commit when I get it fixed and add a comment here about it.

pristinesource:
Actually I am just going to close this pull request

PR #1205, opened 2016-03-25
Improvements to LogDownloadMavLink dialog.

clovett:
Improve LogDownloadMavLink  layout.  Show progress bar while downloading logs.  Fix naming of log files so they match the names shown in the pre-fetched list.  Disable buttons while downloading.  Fixed some other minor bugs (a null ref and some missing Dispose calls).


clovett:
wait a second, there's a bug...

PR #1207, opened 2016-03-25
Fix bugs in LogDownloadMavLink

clovett:
Dispose unused SerialPort objects.  Unregister SerialPort Progress event handler in all cases.  Null ref in MainV2 when switching views quickly.  Localization of strings in LogDownloadMavLink form.  Use TableLayoutPanel for better localizability and stretching behavior.  Unify DownloadThread code.  Add progress bar showing log download progress.  Add helper methods to cleanup some common code like AppendSerialLog.  Fix newlines in log so text flows properly when form is resized.  Remove unused function chk_droneshare_CheckedChanged.


meee1:
just one note for the future, can you make separate changes in separate commits. this way when I review the changes I can see the changes that are required for a single function. at the moment there are multiple changes that cross the same areas.

clovett:
Yeah, sorry about that, in this case I had to checkin a fix.  I have one more related change that makes the dialog layout even more compact, eliminating the blank space in the middle by having 2 columns of buttons but I'll wait till this is integrated first...

clovett:
Since there are no conflicts I will submit a new pull request for these changes.

PR #1209, opened 2016-03-30
Added CameraJoystick

pristinesource:
Added a CameraJoystick class so enable a second joystick to be added to control camera operations (MPO). This MPO control has safety features that are configurable to only allow it to override the main controller (AVO) if it hits a certain threshold. If that threshold is set to 0, then the MPO has 100% control and the AVO has 0% over that channel.


pristinesource:
This is the updated one with all of the issues fixed and I rebased your master branch.

meee1:
i like the change, but your commit history is a little messy. it makes it very hard to track. and would require a bit of time from me to fix. could you possibly see if you can fix this?

pristinesource:
I agree, I had some issues with a rebase.  I will re-fork your repo and apply the updates again into a single commit so it is a lot cleaner.

samworthington:
I'd find this really useful, how is it progressing?

Sam

pristinesource:
Sam,

I have been super busy with work and buying a house lately, but I still 
plan on adding this, I just need to find some spare time.


Sorry for the delay :(

--Anthony

samworthington:
Hi Anthony,

No need to apologise, thanks for the update. We use [vjoy ](http://vjoystick.sourceforge.net/site/index.php/download-a-install/72-download) and [UJR ](http://evilc.com/joomla/articles/9-code/autohotkey/1-ujr-universal-joystick-remapper) to combine multiple joysticks into one at the moment so we can do what we need to but would rather not rely on extra bits of software.

 Good luck with the house and I look forward to this being added.

Sam

pristinesource:
I re-created a new branch for this, so I am going to close this one and start a new pull request.

PR #1215, opened 2016-04-06
Fix bugs in LogDownloadMavLink dialog

clovett:
Fix bugs in LogDownloadMavLink dialog.  Dispose SerialPort Streams properly.  Fix null reference in MainV2 form.

### DIFF
--- a/Log/LogDownloadMavLink.Designer.cs
+++ b/Log/LogDownloadMavLink.Designer.cs
@@ -28,23 +28,35 @@
         /// </summary>
         private void InitializeComponent()
         {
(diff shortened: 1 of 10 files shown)


meee1:
Merged,

I split the commit up into smaller parts, to better show how it should be done. each commit is now independent.

clovett:
thanks, the only reason I didn't do that is because merging your fork back to mine isn't working very well...so doing multiple smaller things is painful... remember how bad the merged looked before? The only way I could find to fix that was to delete my fork and start over.

PR #1266, opened 2016-05-28
Update FlightPlanner.cs

jy723:
sometimes I drag the currentMarker ，the map cann't be dragged when  use mouse left button;


magicrub:
It's best to go ahead and remove code instead of commenting it out like this.

PR #1272, opened 2016-06-01
L10N: auto switch default map, update zh-hans

SkiTiSu:
Thanks.

### DIFF
--- a/GCSViews/FlightPlanner.cs
+++ b/GCSViews/FlightPlanner.cs
@@ -539,6 +539,24 @@ namespace MissionPlanner.GCSViews
                 {
                 }
             }
+            else
+            {
+                if (L10N.ConfigLang.IsChildOf(CultureInfo.GetCultureInfo("zh-Hans")))
+                {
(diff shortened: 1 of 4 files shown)


bys1123:
Thanks

GCSViews/FlightPlanner.cs
```diff
@@ -539,6 +539,24 @@ public FlightPlanner()
                 {
                 }
             }
+            else
+            {
+                if (L10N.ConfigLang.IsChildOf(CultureInfo.GetCultureInfo("zh-Hans")))
+                {
+                    CustomMessageBox.Show("亲爱的中国用户，为保证地图使用正常，已为您将默认地图自动切换到具有中国特色的【谷歌中国卫星地图】！\r\n与默认【谷歌卫星地图】的区别：使用.cn服务器，加入火星坐标修正\r\n如果您所在的地区仍然无法使用，天书同时推荐必应或高德地图，其它地图由于没有加入坐标修正功能，为确保飞行安全，请谨慎选择", "默认地图已被切换");
```
meee1: this will display on every run. i dont think this is a good idea.

SkiTiSu: @meee1 Thank you for your review. After the first time the map set to GoogleChinaMap, the config 'MapType' will save this parameter so it won't be shown next time.

meee1:
merged thanks

PR #1274, opened 2016-06-02
CurrentState: fix battery2 mavlink packet scaling

magicrub:
Battery2 mavlink packet units is same as battery1 units: volts(mV) and current (10*mA). Batt1 was scaled correctly, batt2 had no scaling.


magicrub:
Here's the important arduPilot commit regarding this: https://github.com/ArduPilot/ardupilot/commit/608fbee867384bf2e34ac8c4c2f53fe4ab8ab5db

meee1:
merged, thanks

PR #1320, opened 2016-07-30
updated camerasBuiltin.xml to add gopro hero 4

ricitron:
Added gopro hero 4 wide to builtin cameras for mission planner.


meee1:
merged thanks

Issue #1297, opened 2016-06-24
Add firmware upload for Bebop2


rmackay9:
It would be good to add support for uploading firmware to the Bebop2.

Instructions are in the "Uploading the firmware" section of this wiki page: http://ardupilot.org/dev/docs/building-for-bebop-2.html

There are some tricky bits:
- adb needs to be installed on the user's machine
- we need to modify some scripts on the bebop2 and we want to be able to backout those changes as well so that the default bebop2 firmware can run.

@OXINARF had some ideas about how this could be done.  I've got a Bebop2 and so does Leonard so we'd be happy to test.  If we're going to do this it may be possible to get Parrot (maybe we could even convince @jberaud to send a Bebop2 to the developer who takes this on?).

jberaud:
Yep OK to send a Bebop for this. It's going to be needed for Disco also at some point.
adb is needed. There are other solutions regarding the startup script though they would all involve modifying the startup scripts via adb.

OXINARF:
I could take the job but I can only commit to work on it at the end of July.

meee1:
couple of questions.
1. is the bebop pingable as soon as the wifi is connected?
2. is port 9050 open if adb is not running?

jberaud:
1. yes
2. What do you mean by "open" ?

jberaud:
provided a program is listening on the bebop's side yes but there is nothing listening on that port apart from adb.

meee1:
so I assume the power 4 times trick just starts adb? its not just a firewall mod or something similar.

ie just trying to establish when the bebop is connected, and will accept a adb connection, without actually starting a adb session and waiting for the timeout.

jberaud:
>so I assume the power 4 times trick just starts adb? its not just a firewall mod or something similar.

True
>ie just trying to establish when the bebop is connected, and will accept a adb connection, without actually starting a adb session and waiting for the timeout.

what can be done is connecting the bebop via usb. When pressing the button 4 times, you could just wait for an rndis interface to be present and the DHCP to have succeded, then adb is available.

rmackay9:
oh wow.  I didn't even notice that there was a small usb connector on the back of the Bebop2 right below the red button.  If connecting that would make it easier for @meee1 to implement the firmware upload, from a user point of view that would be a-ok (i.e. no need for it to be over wifi really).

meee1:
@rmackay9 
can you test this via usb (i asume you have the adb usb driver already installed)

[...]

PR #1337, opened 2016-08-19
Add support for uploading firmware to Bebop2

OXINARF:
As talked in #1297 I implemented the firmware upload to the Bebop2 (and it should be mostly ready for Bebop and Disco if they use the same method). It took me longer than expected because of unexpected lack of time and a hard time trying to make detection better.

The detection of the board is where this is the worse. Right now, in MP you first select the firmware to upload and then it tries to detect a connected board. There isn't a good way to detect Bebop2 so my initial idea was to present the user with a list of boards and let him choose the board (showing the detected ones with some kind of mark). I tried to do this but it wasn't working correctly, I wasn't being able to correct it, the code was a mess and I was tired of fighting with it - so I slightly changed my approach: instead of a screen selection it tries to detect the board like it does now but, if it doesn't detect any board, then it asks if it is a Linux board and, if the user says yes, it asks if it is a Bebop2. This is obviously not scalable but it works.

I haven't yet looked at changing the Wizard, so for now, you can only use the Install firmware screen. The binaries automatically built by the autotest server don't work now because they aren't built statically - I will be revising the build_binaries script and then add the correct links for MP to download the firmware. For now, you can build the firmware statically and then upload via Custom Firmware (I've changed that window to allow all files to be viewed since firmware for Linux isn't supposed to have any extension).

@meee1:
- I only noticed now that your commits have some kind of structure in the commit message. If you want me to change my commit messages just tell me what the structure is and I'll change the messages.
- With this we need to ship the adb.exe and AdbWinApi.dll files (included in these commits). I've made it copy these files to the output directory but I don't know if that is enough

@jberaud:
- I hope this is what you where looking for and if there is something that you think that should be changed or added, go ahead and tell me. Sorry it took me this long.
- In the end I reboot Bebop2 by executing the file reboot.sh, is that OK?

@rmackay9: Do you want to test it? I've used it several times, trying to go through all possible code routes but that's not always easy. Also being the developer of this code I'm used to how it works, having an user like you try it, it would be great.

As a final note, two things I'd like comments on:
- if the init script is going to be changed this always does a backup of it. Is this something we want or if we already have a backup we shouldn't overwrite it?
- I've made it that if the ArduCopter line is present in the init script but doesn't match the line in MP then it is replaced - the reason for this was to allow us to update the line in MP (for example, new or different options passed to ArduCopter), but it can be inconvenient if you have personalized that line (changing MAVLink IPs or ports for example).


rmackay9:
I'm very happy to test this is someone can give me a binary or something.

Re the init script.. my guess is that we're safe if we keep a backup of any script that we think came from Parrot whether that's the original script that came before we loaded ardupilot or an updated script if Parrot happens to do a software update on the device.  Otherwise I think we're safe to replace the script that's there without making a backup because I don't think we have any way of know that what was there before was working.  You can imagine if an upload fails to work (and it's because the startup script didn't work) that the user would try the upload again a few times and it wouldn't be good if that zapped all record of the original script.

Anyway, not sure I'm catching the important details but in any case, this is great!  TBH, I wasn't actually expecting it to be done so soon. :-)

OXINARF:
@rmackay9 I'm not sure how @meee1 does MP releases, but here you go with a zip folder (and one working but not flyed Bebop firmware) that should work: https://1drv.ms/u/s!Ai-2ZIA-TjyCpiyicHnLXkHI49ym
My only instruction: use it like you were going to install firmware to any other board (except you don't connect it to your computer). And remember you need to use the custom firmware upload for now.

Regarding the init script: I think it is highly unlikely that a corruption would occur, but I think that your proposal is good. I believe I can detect if it is an original from Parrot or not (although a user could fool the detection).

rmackay9:
@OXINARF, ok, I tried the upload and it seems to work partially (which ain't bad!).
1. I first connected via Wifi to the Bebop from my PC.  Then I selected "AUTO" on the MP's connection drop-down.
2. From MP's firmware install screen, if I selected "beta" or "latest" firmware (i.e. clicked "beta firmwares" link or pushed Ctrl-Q) it seemed to fail while looking for the appropriate firmware.
bebopuploadfailed
3. To get around the above, I directly downloaded the bebop2 firmware to a local directory from here: http://firmware.ardupilot.org/Copter/latest/bebop-quad/
4. , then I clicked on the "Load custom firmware" link and looked for the arducopter-quad binary (it wasn't visible so I changed the window's "Filename" field to "*".  I selected the file and it asked me, "Is this a Linux board?" and then "Is this a Bebop2?" and it didn't work but asked me to push the red button 4 times, and I did and then it seemed to upload and reboot the bebop!  The MP screen seemed to leave "Rebooting..." written at the bottom of the screen but that's no big deal.
5. Next I reconnected to the Bebop2 wifi again ('cuz it had rebooted) and attempted to connect with UDP (port=14550) but it never seemed to successfully connect.  I wonder if maybe the startup script isn't quite right?
Now, the

OXINARF:
@rmackay9 I knew that first long text was a bad idea :smiley: 

1. I still haven't made a PR to add Bebop to the downloaded firmware because:
2. the firmware we are compiling in autotest (the one you downloaded) isn't built statically so it doesn't work

So to test it, you either compile it yourself or can you use the arducopter file I put in the link above.

>  I selected the file and it asked me, "Is this a Linux board?" and then "Is this a Bebop2?" and it didn't work but asked me to push the red button 4 times

What do you mean by this "it didn't work"? Pushing the power button 4 times is a requirement to get adb running and there's nothing we can do to automate it.

> The MP screen seemed to leave "Rebooting..." 

Yes, I can change that, but you got a window saying "Firmware installed" right?

Have you finished your post? It ends with "Now, the"...

@meee1 I think this has been merged right? I will probably need to do a couple of changes but I will open another PR with them.

rmackay9:
Right ok.  I also just now realised that the binary from the centralised server doesn't actually work even if renamed correctly.  If I used an "upload custom firmware" with a known working binary it was ok.

"It didn't work" means that I hadn't pushed the button 4 times.  After I did it was fine.  So that's not a real problem.  I just wrote down my experience for completeness.

Ah, forget the "Now the".. just a cut-paste issue.

rmackay9:
By the way, if you don't do it already as part of the upload, I think we should do a "adb shell chmod a+x /usr/bin/arducopter" to make sure the file is executable.

OXINARF:
That is done as part of the upload, yes.
So basically, it worked, the only thing missing is being able to click the firmware you want without using the custom firmware option.
Regarding this, it was interesting that you connected to the Bebop's Wifi before choosing the firmware - it wouldn't work anyway because you then wouldn't have Internet to download it. So I'll look to add a warning when we can't download a firmware asking if the user is connected to the Internet.

rmackay9:
Ok, txs.  I was on my desktop machine though which has an ethernet connection and it seems to stay connected to the internet even if I connect to the Bebop2 through wifi.
Really excited about this feature!

OXINARF:
@rmackay9 Yes, that is true, if you have Ethernet you should stay connected to the Internet. That was actually an excellent testing setup.
I will try to work on the improvements for MP this week. To make everything work faster I'll probably look at getting the firmware built statically before rewriting the build_binaries script.

meee1:
this is all in now and working.

rmackay9:
Yes, it works great!

PR #1380, opened 2016-10-04
Explain possible stand-alone use of LogAnalyzer

kralo:
LogAnalyzer is also a great tool if you're on linux but I wasn't aware that you can use it without installing anything of MissionPlanner, so a quick note to others :smile:


meee1:
i think this pull request would be better suited for here
https://github.com/ArduPilot/ardupilot/tree/master/Tools/LogAnalyzer

PR #1387, opened 2016-10-15
Radio: fixed binding with DSM-X and DSM-8

tridge:
should be in param1 not param2


tridge:
hmm, actually, common.xml says param2, I'll change ardupilot to accept either

Issue #1388, opened 2016-10-15
Joystick config does not save or load properly


nanthony21:
I can see that the joystickbuttons.xml and joystickaxis.xml files save properly in User/Documents/Mission Planner but when the configuration window is opened it is always completely unset, just like the first time it was opened. 

It also seems that the .xml files are only properly saved to if the "autodetect" feature isn't used.

meee1:
are you connected to anything when you config the joystick? MP stores separate joystick configs for plane and copter.

gediminasgu:
I can confirm that Joystick panel doesn't show saved values, even seems like mapping is stored correctly. See the video: http://d.pr/i/uqTo
MP version: 1.3.41 build 1.1.6110.31287

lakshaydang:
I am also not able to save the values that i configured for my joystick. Help me !

PR #1397, opened 2016-10-23
Fix for joystick setup file load issue #1388

gediminasgu:
Previously joystick file did try to load from local folder, which is not where setup files are stored, that's why joystick setup wasn't loaded or saved.


gediminasgu:
Fix for #1388

PR #1416, opened 2016-11-28
Improved theming functionality.

mredstall:
Mission Planner: Improvements to theming.

Fixed a number of instances where themes were not applied correctly to controls.  Changed the boolean option for 'advanced layout' to support multiple layout configurations instead.  Other general improvements to theming.

This affects a large number of files, but frequently this is just a single line: adding a call to the theme manager to apply the theme to the control.


meee1:
As you have seen I partial merged with a few changes.
things like the mybutton colours where not complete. and still need to add a check if the current config is advanced, to change the layout to advanced. so users see no change

mredstall:
I will change the button colours for mouse over and mouse down to match the
current release. (were there any other inconsistencies that need to be
fixed?).

Do you have a recommended approach to transitioning the advanced view
boolean (check if the boolean setting exists and if so read its value and
then delete it?).

meee1:
Ive already made the advanced view change
559ee77f1c392d648a20ab2a7531f51f480b2f44
as i had several people report missing buttons in the beta.

3c0bbb83bf046354cb11abaa13f7ea661fddd3d3 this is also the change about thememanger, apply to all child windows, to prevent having to add code to all children forms. Either inheritance of this change will apply themes.

meee1:
closing this pull, as other one is newer

PR #1421, opened 2016-12-01
Further theming improvements.

edwinhayes:
Theming now correctly affects waypoint and action panes on flight planning tab.  Fixed issue when attempting to load an invalid theme.  Added options for theming terminal screen.  Additional layout configuration options.

This pull request comes from a separate branch than #1416 , due to some git insanity at our end.


edwinhayes:
Also, apparently I changed out of mredstall's account at some point...

meee1:
can you explain the pupose of terminaltheming? ie what does it actually do/intended to do?

edwinhayes:
Noticed that the on the (now seldom used) terminal tab, the actual terminal part was using the same theme colours as the rest of the interface: the option allows using different bg/fg colours for the terminal, so you can have white text of black, even in light coloured themes.

Not really sure why Michael (our intern, who actually made the changes) created a checkbox to enable it though, it could just be set as part of the actual theme.

mredstall:
I have removed the terminal theming checkbox and used a boolean variable inside the theme manager instead. It didn't really make sense for users to change this.

meee1:
this has been merged. however a few things have been changed, as they break existing functions.

PR #1463, opened 2017-01-22
LogBrowse: fix the display of log row data (mono)

tatsuy:
Log data doesn't display in DataGridView1.
I reverted the getting of row data to the commit: 3b542617.


meee1:
i made changes yesterday for this. so this is no longer needed/required

tatsuy:
Thank you! I'll check it.

PR #1473, opened 2017-02-07
Feature: Add airdistance to Flight Plan

patschn:
I have added a column for the airdistance in the FlightPlan. It would be nice to see not only the grounddistance, but also the airdistance (contains the difference of the altitude) in the flightplan. 
The distance is being calculated with the help of the SharpKML-MathHelpers library.

So a flight plan could look like this:

screenshotdistgdista

Modifications:
- The DataGridView contains an new column (DistA for airdistance and DistG, which was the column Dist)
- setgradanddistandaz() calculates the distance with the SharpKML-Library (see double distancexyz = ...)


magicrub:
Is the sharp KML library already used or is it a new dependency? If it's new, i suggest we remove it because this is just a simple Pythagorean that cloud be done manually quite easily.

patschn:
The SharpKML was already included in the MissionPlanner solution. I didn't include a new reference. Of course I can change this calculation to a pythagorean (see 'First simple implementation', commit: 54848c6 in feature/XYZDistance), if requested.

patschn:
Fixed merge problem after commit e33c7c125c9fdc2f13243d12cf771041937115ec from MP-Master

patschn:
@magicrub @meee1 Any other concerns with this PR? I'm open for discussion.

meee1:
my current concern is the fact there are 6 commits for something I would only really concider needing 1. as it only touch's flightplanner. ive got no issue with the idea. just currently I looks very messy

magicrub:
up to @meee1 .

patschn:
@meee1, @magicrub I understand your concerns.
The last commit was not definately not planned. A fix was required for one of the last commits of MP-master and Git somehow suggested a merge. 
Well I can, of course, create a new branch based on top of the last tag of MP, with a clean commit and open a new pull request and implement it as shown in the picture above with the SharpKML library.

But would the idea itself be accepted?

patschn:
The version of the dll-reference seems to change with building the solution (MP 1.3.44). For example: Is it possible to prevent this change? I find no way to prevent this at the moment. 

May I ask: Are the changes in FlightPlanner.cs, flightplanner.resx and FlightPlanner.Designer.cs considered as messy? Because the changes in the last two files were just Visual Studio changes.

Thank you in advance

meee1:
ive updated Dist to be 3d distance
https://github.com/ArduPilot/MissionPlanner/commit/9b7e53acf92837dca80596a3151e0f12391ef075

meee1:
the pull request has changes for
https://github.com/ArduPilot/MissionPlanner/pull/1473/commits/2eca24b5872e29e7eb829f13e8762272a3cb73cc

which are unrelated to the original intent of the pull request. this is what i mean by messy

patschn:
Ok, that clears it up. Thank you! [Pull  Request can be closed if there are no other concerns]

PR #1521, opened 2017-04-15
Fixed bug that would cause drift in the joystick

ExusAltimus:
Fixed bug that would cause drift in the joystick if the RC Channel's Min / Max are anything but 1000 and 2000

### DIFF
--- a/Joystick/Joystick.cs
+++ b/Joystick/Joystick.cs
@@ -1447,7 +1447,7 @@ namespace MissionPlanner.Joystick
                     break;
             }
             // between 0 and 65535 - convert to int -500 to 500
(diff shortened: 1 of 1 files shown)


ExusAltimus:
I was having extreme drift issues with a Logitech F310 for days, even after configuring the deadzones and recentering multiple times / testing with multiple controllers. Turns out the math for the joystick only works if the PWM Min / Max are 1000 and 2000 because the range ends up being 1000, and for some reason the joystick is mapped to -500 and 500?

meee1:
this is now fixed, using a map functions to ensure it scales as desired

PR #1532, opened 2017-04-27
MAVState.cs: array size to match extended ID space

EShamaev:
Size of packetspersecond and packetspersecondbuild arrays was 256 and during receive of Mavlink v2 messages exception was generated in MAVLinkInterfac.cs at refernce to packetspersecond[msgid].


meee1:
Thanks for this. very much a bug in mavlink 2 receiving. I modified how it was done to use a dictionary vs an array, the same as packets bellow it. but credited you in the commit

EShamaev:
Dictionary is great, too bad I did not think of that myself.

Issue #1422, opened 2016-12-05
Show only onboard compass calibration when vehicle has the capability


OXINARF:
With https://github.com/ArduPilot/ardupilot/pull/5325 merged to master, I'd love to see the live calibration be hidden away and only show the onboard one (maybe with a button to show the live calibration if a user really wants it).

proficnc:
Please +1 to this... people appear to think these are both the same... but they are not.. 

for Pixhawk 2.1, i would prefer people only used the onboard calibration... Thanks

OXINARF:
Fixed with c409c1a5f3f7c678e3edf03dc4f27a3bcfabee7c and cfed796bee77143d784c34ce69c1b7fe0a854226

PR #1550, opened 2017-05-11
ConfigHWCompass: show both options if autopilot doesn't present onboard capability

OXINARF:
Issue #1422 was implemented in c409c1a5f3f7c678e3edf03dc4f27a3bcfabee7c by showing the onboard or the live calibration if the compass calibration capability is present or not.

Unfortunately ArduPilot didn't include this capability when it started supporting the onboard calibration, notably Copter before 3.4.4 and the current Plane 3.7.1 release don't have it. This has led to complaints, especially from Plane users, that they can't use the onboard calibration.

This change makes only the onboard calibration available when the capability is present but show both types of calibration when it isn't.


OXINARF:
Done in a different way in cfed796bee77143d784c34ce69c1b7fe0a854226

PR #1664, opened 2017-10-16
Updating required Visual Studio version note

Naphier:
Changing from VS Express 2013 to VS Community 2017. I tried opening in VS2015 and it failed reading the slns, so I went through and downloaded VS2013 to see if that would work. An hour later I found out that I wasted my time and could just open in 2017...

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ How to compile
 
 3. Build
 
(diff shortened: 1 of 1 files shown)


Naphier:
After finding more clues as to potential issues when setting up project I'll be submitting a more thorough pr so others don't fall into the same trap.

PR #1671, opened 2017-10-20
Add px4fmu-v4pro support

klopezal:
This pull request adds Pixhawk 3 Pro support to Mission Planner. The corresponding firmware is available in a stable branch for ArduPlane, ArduCopter and ArduRover at the moment. Another PR has been submitted in order to update the XML firmware file : https://github.com/ArduPilot/binary/pull/9

### DIFF
--- a/Utilities/Firmware.cs
+++ b/Utilities/Firmware.cs
@@ -272,7 +272,7 @@ namespace MissionPlanner.Utilities
                                     temp.urlpx4v2 = px4v2;
                                     temp.urlpx4v3 = px4v3;
                                     temp.urlpx4v4 = px4v4;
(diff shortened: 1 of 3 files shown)


Utilities/Firmware.cs
```diff
@@ -267,6 +272,7 @@ public List<software> getFWList(string firmwareurl = "")
                                     temp.urlpx4v2 = px4v2;
                                     temp.urlpx4v3 = px4v3;
                                     temp.urlpx4v4 = px4v4;
+                                    temp.urlpx4v4 = px4v4pro;
```
meee1: bug

klopezal: Sorry for missing it, it is now fixed.

Issue #1682, opened 2017-11-10
Menu Bar disappears when disabling autohide when autohide is enabled in config file


JWangAeronavics:
#### Issue details
When mission planner is started up with the menu bar autohide feature enabled in the saved config file, disabling autohide causes the menu bar to disappear as soon as the mouse cursor is moved away from the menu bar until mission planner is restarted.

This is being caused by a double registration of the MainMenu_MouseLeave() event handler on startup, once when MainV2 is initialised, and a second time when AutoHideMenu() is called whilst loading from the config file.

#### Version
1.3.50

#### Platform
[x] All
[  ] AntennaTracker
[  ] Copter
[  ] Plane
[  ] Rover
[  ] Sub

![image](https://user-images.githubusercontent.com/32720330/32637139-fc1d0a38-c61c-11e7-9335-f9ff64de0b90.png)

JWangAeronavics:
possibly the same issue as #4 which was closed in 2013

PR #1683, opened 2017-11-10
Fixed menu bar disappearing when autohide is disabled

JWangAeronavics:
This fixes the double event handler registration issue described in issue #1682


edwinhayes:
@JWangAeronavics, is this still broken in master?

JWangAeronavics:
@edwinhayes the problem still exists in master

meee1:
there is a other underlying issue here that has been created.
if a user wants to enable autohide, with the current structure they cannot, as it is forced by the displayview.

JWangAeronavics:
@meee1 sorry that's my bad, the above commit fixes that issue.

meee1:
can you actually tell me how the original issue is created? as I cant replicated the underlying issue.

JWangAeronavics:
Steps to reproduce:
1. turn on autohide 
2. quit mission planner and ensure the menu_autohide config is set in config.xml
3. restart mission planner
4. disable autohide
5. mouse over the menubar then move the cursor off the menubar 

this assumes there is no theme that forces autohide present

meee1:
I still cant replicate this issue. this feature is broken however as the setting is not remembered between sessions.
my test was with no displayview defined in settings at all. and as per your instructions.

JWangAeronavics:
can you repeat this test either with e9107bf5d79eb5fbdc8edff2031bc39f67459fbe cherry picked in or the autohide displayview change #1684  reverted?

PR #1684, opened 2017-11-10
Added the ability to create layouts with menu autohide always turned on.

JWangAeronavics:
Added the ability to create custom layouts where the menu is always hidden by default to save on screen space. The ability to toggle the autohide feature is also hidden from layouts with this feature turned on.
This feature has not been applied to any of the included layouts.


edwinhayes:
@JWangAeronavics can you address these conflicts at some point?

PR #1686, opened 2017-11-21
MavLink interface enhancements to improve handling of non-fc components

JWangAeronavics:
When mission planner is connected to a system with many MavLink enabled components, it does not connect to a component in a deterministic manner and may automatically connect to a component that is not directly controllable (such as a UART bridge). It is also impossible to identify which component is which as the component ID is not included in the connection selection combobox. 

These changes forces mission planner to look for the root component of a system (the flight controller) and connects to it automatically when opening a connection and cleans up the enumeration of the slave components by identifying them by component ID (or enumerated name if available). They also stop mission planner from continuously switching between sysid/compid combinations on initial connection and fix some buggy behaviour when switching between components without forcing a refresh from the full parameter list.

before (note the lack of flight controller data on connect):
connectionbefore

after:
connectionafter


Controls/ConnectionControl.cs
```diff
@@ -130,21 +130,40 @@ private void CMB_sysid_SelectedIndexChanged(object sender, EventArgs e)
                     MainV2.comPort.sysidcurrent = temp.sysid;
                     MainV2.comPort.compidcurrent = temp.compid;
 
-                    if (MainV2.comPort.MAV.param.Count == 0 && !(Control.ModifierKeys == Keys.Control))
-                        MainV2.comPort.getParamList();
+                    //force param update and reload view to prevent missing information in config pages
+                    MainV2.comPort.getParamList();
+                    MainV2.View.Reload();
```
meee1: this change implies that all nodes support the param protocol. which is why it was optional

Controls/ConnectionControl.cs
```diff
@@ -130,21 +130,40 @@ private void CMB_sysid_SelectedIndexChanged(object sender, EventArgs e)
...
+                    {
+                        //use name from enum if it exists, use the component ID otherwise
+                        mavComponentString = compid.ToString();
+                        if (mavComponentString.Length > mavComponentHeader.Length)
+                        {
+                            //remove "MAV_COMP_ID_" header
+                            mavComponentString = mavComponentString.Remove(0, mavComponentHeader.Length);
+                        }
+                    }
+                    e.Value = ((int)temp.sysid) + "-" + mavComponentString.Replace("_"," ");
```
meee1: there is no distiction on which physical interface this is comming in on. ie some people use 2 links via diffrent methods to the same vechile.

Controls/ConnectionControl.cs
JWangAeronavics: The original code does not account for this either. It only updates if the param list if the current param count is 0. Currently any device that does not support the param protocol will spin forever, trying to fetch parameters unless interrupted by the user

meee1: also im against a full refresh on changing the target node. if the list has already been received

Mavlink/MAVLinkInterface.cs
```diff
@@ -464,61 +463,31 @@ private void OpenBg(object PRsender, bool getparams, ProgressWorkerEventArgs pro
 
                     Thread.Sleep(1);
 
-                    // can see 2 heartbeat packets at any time, and will connect - was one after the other
+                    // listen for heartbeats from component id 1 (flight controller) only ensures that at least one flight controller is operational
```
meee1: if you are connecting straight to a gimbal, without a flight controler whats the expected result?

Controls/ConnectionControl.cs
JWangAeronavics: Is this currently possible? How would I go about testing this? Would adding the port back into the display string resolve this?

Mavlink/MAVLinkInterface.cs
JWangAeronavics: The expected result would be that it times out. What would be the use case of connecting a gimbal without a flight controller?

Controls/ConnectionControl.cs
JWangAeronavics: I've dug deeper into this and it looks like there is an issue where if parameters are refreshed on a component that does not have apname set to ARDUPILOTMEGA is causing parameter corruption in mission planner on components with apnames set to ARDUPILOTMEGA. Will have to investigate this further...

Controls/ConnectionControl.cs
```diff
@@ -131,20 +131,40 @@ private void CMB_sysid_SelectedIndexChanged(object sender, EventArgs e)
                     MainV2.comPort.compidcurrent = temp.compid;
 
                     if (MainV2.comPort.MAV.param.Count == 0 && !(Control.ModifierKeys == Keys.Control))
-                        MainV2.comPort.getParamList();
+                       MainV2.comPort.getParamList();
```
JWangAeronavics: I've reverted this change and fixed the root cause of the missing data issue

Controls/ConnectionControl.cs
edwinhayes: @meee1 could you give us some feedback so that we can determine how best to proceed with the work we're currently doing?

Controls/ConnectionControl.cs
meee1: nice spot on this one. the getInfoFromStream is definitely a bug.
the SetupMavConnect is not technically a bug, as the MAV reference is set inside the same function. but for clarity its better.

meee1:
I just noticed I had another issue around the connect. so I changed it here. #1698 
I will try to commit your second change here

meee1:
ive done a part merge of the seconds commit. fixing the param bug

Controls/ConnectionControl.cs
edwinhayes: @JWangAeronavics, in lieu of feedback from @meee1, can we just update the string to contain something identifying the physical port?  "S1C1 (ARDUPILOT) on UDP:14550" and "S1C57 (UART_BRIDGE) on COM6" perhaps.

Mavlink/MAVLinkInterface.cs
edwinhayes: @JWangAeronavics, in view of other changes which you mentioned @meee1 had made to handling components other than the flight controller, how much of this is still relevant?

JWangAeronavics:
@meee1 I've merged in your changes for #1698 and although it fixes the 1 HB issue as described the automatic target selection after polling for heartbeats still behaves indeterminately. The method getHeartBeat overrides the target sysid/compid every time it is called and receives a heartbeat and there is at least one instance of it being called after the OpenBg method tries to set the target to a flight controller's sysid/compid.

JWangAeronavics:
I've re-implemented the changes that specifically deal with the target sysid/compid being rewritten on getHeartBeat() on top of the changes for #1698 in this pull request

Controls/ConnectionControl.cs
```diff
@@ -164,7 +164,7 @@ private void cmb_sysid_Format(object sender, ListControlConvertEventArgs e)
                             mavComponentString = mavComponentString.Remove(0, mavComponentHeader.Length);
                         }
                     }
-                    e.Value = ((int)temp.sysid) + "-" + mavComponentString.Replace("_"," ");
+                    e.Value = ((int)temp.sysid) + "-" + mavComponentString.Replace("_"," ") + " on " + temp.port.BaseStream.PortName;
```
meee1: any reason for this ordering?
ie when trying to find something having the broadest to narrowest makes sense. ie a drill down approach

this is currently adhoc

broadest being port then sysid then component

edwinhayes: This depends on which use-case you believe is more common: selecting between multiple systems (or components) communicating on a single port, or selecting between multiple ports connected to a single system.

In my opinion, selecting between ports shouldn't even be exposed to the user in this manner: if an aircraft has multiple communications links, they should be aggregated together, rather than treating them separately (refer #1481 ).

JWangAeronavics:
I have changed the display string back to <port>-<system ID>-<Component>

PR #1689, opened 2017-11-26
Added sysid on each marker when >1 vehicle connected

jabelone:
When only one vehicle is connected there are no visible changes. I've introduced a small check in the "draw all mavs" loop. If there is more than one vehicle connected it displays the sysid below the altitude. At the moment it's very difficult to determine which vehicles are which and this makes it much clearer. 

Screenshot:
image

Apologies if I've done something wrong as I have very little idea about Mission Planners internal workings. (or C#/.NET for that matter)


meee1:
I'm not going to merge this one as it is
best place for this mod is in common.cs
in function getMAVMarker

PR #1692, opened 2017-11-27
Expanded on the configuration options of the "display view" feature

JWangAeronavics:
These changes expand on the existing display view feature for creating custom layouts and allow developers to have finer control over what configuration options are available to the end user.

### DIFF
--- a/MainV2.cs
+++ b/MainV2.cs
@@ -514,6 +514,12 @@ namespace MissionPlanner
                 }
                 t.SelectedIndex = 0;
             }
+
+            if (MainV2.instance.FlightPlanner != null)
+            {
+                //hide menu items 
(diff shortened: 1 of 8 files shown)


MainV2.cs
```diff
@@ -417,12 +417,29 @@ public void updateLayout(object sender, EventArgs e)
             MenuSimulation.Visible = DisplayConfiguration.displaySimulation;
             MenuTerminal.Visible = DisplayConfiguration.displayTerminal;
             MenuHelp.Visible = DisplayConfiguration.displayHelp;
+            MenuDonate.Visible = DisplayConfiguration.displayDonate;
```
meee1: i will let this through for now.

PR #1704, opened 2017-12-12
Added ability to commit params to flash in param list/tree

JWangAeronavics:
also fixed ctrl+Y command so that it sends a write params command
rather than a read params command

This change is to support components other than flight controllers running ardupilot which require a PREFLIGHT_STORAGE command to save params to non-volatile memory as per the mavlink standard.

### DIFF
--- a/ExtLibs/Utilities/DisplayView.cs
+++ b/ExtLibs/Utilities/DisplayView.cs
@@ -86,6 +86,7 @@ namespace MissionPlanner.Utilities
         public Boolean displayAdvancedParams { get; set; }
         public Boolean displayFullParamList { get; set; }
         public Boolean displayFullParamTree { get; set; }
+        public Boolean displayParamCommitButton { get; set; }
         public Boolean displayBaudCMB { get; set; }
         public Boolean displaySerialPortCMB { get; set; }
         public Boolean standardFlightModesOnly { get; set; }
@@ -164,6 +165,7 @@ namespace MissionPlanner.Utilities
             displayAdvancedParams = false;
             displayFullParamList = false;
             displayFullParamTree = false;
+            displayParamCommitButton = false;
             displayBaudCMB = true;
             standardFlightModesOnly = false;
             displaySerialPortCMB = true;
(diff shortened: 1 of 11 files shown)


meee1:
a better place for this change would be here
https://github.com/ArduPilot/MissionPlanner/blob/master/Mavlink/MAVLinkInterface.cs#L3868
one line

meee1:
means there is no need for a button that would only confuse most users

edwinhayes:
Issuing a store command automatically is not consistent with the original protocol specification, and it's unreasonable to assume that just because a parameter is being changed that the user wishes this change to persist after a reboot.

meee1:
What is your target platform for this change?

edwinhayes:
In addition to flight controller (Pixhawk V1.2 plus AC), our aircraft run a companion computer which runs a number of other MAVLink enabled logical nodes, plus MAVLink compatible gimbals and other microcontroller components which also talk MAVLink.

edwinhayes:
I can ask @JWangAeronavics to allow masking out the additional controls in the layout/theme, so that it's not visible to normal users?  It's my preference to use that approach rather than making the change in a branch of our own, to reduce the number of lines we differ from the mainline.

meee1:
I think this is a function I would prefer to be behind one of the advanced masks.
I just saw your reply. Yes

meee1:
another issue I have with this in general is any screen that changes a param would also need a write button.
hence the 10 seconds delay save.

edwinhayes:
Hmm.  I see how that could be relevant.  It probably wouldn't actually affect us, because (off the top of my head) the other pages are only changing parameters on the flight controller, where it doesn't matter anyway, whereas all our other components will only having parameters changed from the parameters tree/list screens.  How about we add a preference to enable/disable the automated saving behaviour, and potentially mask this off as well.  Then both behaviours are possible.

JWangAeronavics:
changes made as per Edwin's suggestions above

meee1:
can this be rebased? the change history is a little messy at the moment. very hard to track

JWangAeronavics:
the commits have been rebased on top of the master branch changes

PR #1709, opened 2017-12-18
simple interface tab for MavLink Gimbals

JWangAeronavics:
Adds a tab with a series of sliders to the Flight Data page
for testing/controlling gimbals that only operate via mavlink
messages using the mount control interface on ardupilot flight
controllers. This tab will appear when mission planner is
connected to a flight controller and a mavlink gimbal hearbeat is
detected on the same system ID.

image

### DIFF
--- a/GCSViews/FlightData.Designer.cs
+++ b/GCSViews/FlightData.Designer.cs
@@ -45,8 +45,6 @@
             this.BUT_resumemis = new MissionPlanner.Controls.MyButton();
             this.CMB_mountmode = new System.Windows.Forms.ComboBox();
             this.BUT_mountmode = new MissionPlanner.Controls.MyButton();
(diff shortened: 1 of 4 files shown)


Controls/ConnectionControl.cs
```diff
@@ -150,6 +150,12 @@ private void CMB_sysid_SelectedIndexChanged(object sender, EventArgs e)
                     if (MainV2.comPort.MAV.param.Count == 0 && !(Control.ModifierKeys == Keys.Control))
                         MainV2.comPort.getParamList();
 
+                    //enable the payload control page if a mavlink gimbal is detected
```
meee1: please start using the mavchanged event, vs placing code everywhere to do this.

JWangAeronavics: thanks, changed to use the mavchanged event

PR #1728, opened 2018-01-17
changed wp speech

jelkins8:
to say guided and return to launch instead of always "headed to waypoint"

### DIFF
--- a/Common.cs
+++ b/Common.cs
@@ -303,14 +303,15 @@ union px4_custom_mode {
 
         public static string speechConversion(string input)
         {
-            if (MainV2.comPort.MAV.cs.wpno == 0)
-            {
-                input = input.Replace("{wpn}", "Home");
-            }
-            else
-            {
-                input = input.Replace("{wpn}", MainV2.comPort.MAV.cs.wpno.ToString());
-            }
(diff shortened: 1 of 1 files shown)


Common.cs
```diff
@@ -303,14 +303,15 @@ static void chk_CheckStateChanged(object sender, EventArgs e)
 
         public static string speechConversion(string input)
         {
-            if (MainV2.comPort.MAV.cs.wpno == 0)
-            {
-                input = input.Replace("{wpn}", "Home");
```
meee1: you apear to have removed this function here?

jelkins8:
Hadn't realized I removed that, added back.

jelkins8:
Sorry, missed that adding back the home check. Should be good now.

Common.cs
```diff
@@ -303,14 +303,20 @@ static void chk_CheckStateChanged(object sender, EventArgs e)
 
         public static string speechConversion(string input)
         {
+            if (MainV2.comPort.MAV.cs.mode != "Auto" && MainV2.comPort.MAV.cs.mode != "Guided" && MainV2.comPort.MAV.cs.mode != "RTL")
+                input = input.Replace("Heading to Waypoint {wpn}", "");
```
meee1: the reason I haven't committed this is because input is a user defined string, and the replace is over bearing and will only hit one condition which a simple thing like case will invalidate it

PR #1739, opened 2018-01-25
Utilities : fix FFT parsing

BaptisteEspivent:
Fix the problem with FFT parsing


meee1:
i fixed this with globalisation. as your fix was very region specific

PR #1752, opened 2018-02-02
Adding SSH terminal feature

csiriett:
Adding SSH terminal feature to terminal in Mission Planner.
This feature allows a user to SSH to another device using the SSH.net library. The primary purpose of this feature is to allow users to SSH to drones, to make minor changes as required.

Received data from the SSH client is interpreted using various new interpreters, depending on the received ANSI codes. These interpreters make the necessary changes to the terminal to display the same output that would be displayed on the host client. An additional file to interpret Xterm color codes has also been added.

For aesthetic purposes the connect and disconnect buttons on the terminal screen have been merged into a singular button. A text box for the required IP Address when using the SSH terminal has also been added to the terminal screen.


meee1:
ive just merged this change here with a few modifications. mainly separating the logic out
993b49e730853b181f6652a92f7503c097d5b382

PR #1758, opened 2018-02-09
Map overlay

Buckton-Wishart:
Added 5 features to mission planner:

- A km left of travel according to battery life radius display to FlightData. Circle appears around either home base or aircraft

-An Elevation colour gradient display to FlightData and FlightPlanner. Colours go from Red – Yellow – Green – Blue – Purple. Where Red is the highest point on screen and purple the lowest.

-A relative elevation colour gradient display, depending on the altitude and specified ground clearance of the aircraft to FlightData and FlightPlanner. Colours go from Red – Yellow – Green. Where Red means the aircraft will crash and green means you can fly there.

-RF_Propagation interference display as a dotted white line, based on aircrafts altitude and ground clearance. Added to FlightData and FlightPlanner

-Horizon line of sight display as a blue line. Added to FlightData and FlightPlanner

All features are controlled by check buttons and features can be tuned in config tuning, eg adjust the resolution of the elevation display


Naterater:
Suggestion: move the center of the circle to compensate for the wind & trim_arspd_cm. Attached is a commercial aviation solution that is a little more advanced, but would be very neat to integrate.

It's not necessary, but it would add a nice touch to your great features.
image

meee1:
ive added this in a testing form. currently only accessible via control-w

PR #1761, opened 2018-02-10
Uploader: Fix broken logic in PR 1760

pkocmoud:
Uploader: Fix broken logic and added braces to make the nested condition more readable.

Sorry @meee1 the logic in my previous PR was flawed. It is oddly complicated. PR1760 creates a condition where the exception would never be thrown.

This fixes my error and has been thoroughly tested with both good and bad firmware on a PixHawk v1, X2.1 and Pixracer.


meee1:
just to be clear with this.
what will your board report. 
and what will the firmware file be reporting?

meee1:
this should be fixed now.

PR #1816, opened 2018-04-22
Fixed a bit strange part of Japanese display.

tajisoft:
I'm ArduPilot user in Japan.
I found some strange Japanese. So I'll fix it step by step.

### DIFF
--- a/GCSViews/ConfigurationView/ConfigSimplePids.ja-JP.resx
+++ b/GCSViews/ConfigurationView/ConfigSimplePids.ja-JP.resx
@@ -118,6 +118,6 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="TXT_info.Text" xml:space="preserve">
-    <value>注意：このインターフェースを使用するとカスタム設定したPID値がリセットされる可能性があります。</value>
+    <value>注意：このインターフェースを使用するとカスタム設定したPID値の一部がリセットされることがあります。</value>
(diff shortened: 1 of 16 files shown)


GCSViews/ConfigurationView/ConfigSimplePids.ja-JP.resx
```diff
@@ -112,12 +112,12 @@
...
-    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <resheader name="writer">
-    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="TXT_info.Text" xml:space="preserve">
-    <value>注意：このインターフェースを使用するとカスタム設定したPIDがリセットされるかも知れません。</value>
+    <value>注意：このインターフェースを使用するとカスタム設定したPID値がリセットされる可能性があります。</value>
```
muramura: Because possibility「可能性」 is said to be the property of whether it is possible or not. . . I feel like I can do what I want. It tells us that unintended things may happen before change.
Original:
NOTE: using this interface may reset some off your custom pids.

PR #1838, opened 2018-05-21
Mp fixes

davidbuzz:
here's three fixes for MP.  one bugfix for airspeed, one improvement for logo loading, and one commit that colourises icons when flying more than one plane at a time.


Bill-B:
For the airspeed bugfix, should lines 2217 to 2221 also be changed in FlightData.cs?

https://github.com/davidbuzz/MissionPlanner/blob/df315f759e75dc25d67dc2a065871e60986d2826/GCSViews/FlightData.cs#L2217-L2221

PR #1887, opened 2018-07-10
Added Altitude Angel flight reports on the map and flight reporting on arm/disarm

rupertbenbrook:
These changes add Altitude Angel flight reporting to Mission Planner. This includes displaying other flight reports on the map so you are aware of them, and reporting your own flights. To do this we summarise a flight plan to a centroid and radius (or current location if there’s no flight plan), and on arm we submit the flight report, and complete the flight report on disarm. There are also settings that control if this happens or not, and the metadata used for the flight report.


rupertbenbrook:
Build failed, see below. Think some builds need cleaning up in AppVeyor.

Uploading artifacts...
[1/5] ChangeLog.txt (156,687 bytes)...100%
[2/5] checksums.txt (46,443 bytes)...100%
[3/5] MissionPlannerBeta.zip (103,894,169 bytes)...100%
Maximum allowed artifact storage size of 50000 Mb will be exceeded.

rupertbenbrook:
Replaced with PR #1893

PR #1893, opened 2018-07-17
Add Altitude Angel flight reporting to Mission Planner

rupertbenbrook:
These changes add Altitude Angel flight reporting to Mission Planner. This includes displaying other flight reports on the map so you are aware of them, and reporting your own flights. To do this we summarise a flight plan to a centroid and radius (or current location if there’s no flight plan), and on arm we submit the flight report, and complete the flight report on disarm. There are also settings that control if this happens or not, and the metadata used for the flight report.

### DIFF
--- a/Utilities/AltitudeAngel/AltitudeAngel.cs
+++ b/Utilities/AltitudeAngel/AltitudeAngel.cs
@@ -1,32 +1,46 @@
 ﻿using System;
-using System.Text;
-using System.Windows.Forms;
+using System.Threading.Tasks;
 using AltitudeAngelWings;
-using AltitudeAngelWings.ApiClient.Client;
+using AltitudeAngelWings.Extra;
 using AltitudeAngelWings.Service;
-using AltitudeAngelWings.Service.FlightData;
-using AltitudeAngelWings.Service.FlightData.Providers;
-using AltitudeAngelWings.Service.Messaging;
-using MissionPlanner.Controls;
+using MissionPlanner.GCSViews;
 
 namespace MissionPlanner.Utilities.AltitudeAngel
 {
-    public class AltitudeAngel : IDisposable
+    internal static class AltitudeAngel
     {
-        public static MissionPlannerAdaptor MP = new MissionPlannerAdaptor();
-        private static MessagesService Message = new MessagesService();
-
-        public static AltitudeAngelService service = null;
-
-        static AltitudeAngel()
+        internal static void Configure()
         {
-            service = new AltitudeAngelService(Message, MP,
-                new FlightDataService(new ObservableProperty<long>(3), new MissionPlannerFlightDataProvider(null)));
+            AltitudeAngelPlugin.Configure();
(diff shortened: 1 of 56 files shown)


Utilities/AltitudeAngel/AltitudeAngel.cs
```diff
@@ -1,32 +1,46 @@
...
-
-        public static AltitudeAngelService service = null;
+        internal static void Configure()
+        {
+            AltitudeAngelPlugin.Configure();
+            ServiceLocator.Register<IMissionPlanner>(l => new MissionPlannerAdaptor(
+                new MapAdapter(FlightData.instance.gMapControl1),
+                new MapAdapter(FlightPlanner.instance.MainMap),
+                () => FlightPlanner.instance.GetFlightPlanLocations()));
+            ServiceLocator.Register<IMissionPlannerState>(l => new MissionPlannerStateAdapter(
```
meee1: im pretty sure this reference will fail, as its using a class that will be replaced at multiple levels

Utilities/AltitudeAngel/AltitudeAngel.cs
```diff
@@ -1,32 +1,46 @@
...
+                await service.SignInIfAuthenticated();
+                return;
+            }
+            if (CustomMessageBox.Show(
+                    "Do you wish to enable Altitude Angel airspace management data?\nFor more information visit [link;http://www.altitudeangel.com;www.altitudeangel.com]",
+                    "Altitude Angel - Enable", CustomMessageBox.MessageBoxButtons.YesNo) == (int)CustomMessageBox.DialogResult.Yes)
+            {
+                await service.SignInAsync();
+            }
+            settings.CheckEnableAltitudeAngel = true;
```
meee1: please check the user provided input vs just setting true

Utilities/AltitudeAngel/AltitudeAngel.cs
```diff
@@ -1,32 +1,46 @@
...
-            service = new AltitudeAngelService(Message, MP,
-                new FlightDataService(new ObservableProperty<long>(3), new MissionPlannerFlightDataProvider(null)));
+            var settings = ServiceLocator.GetService<ISettings>();
+            var service = ServiceLocator.GetService<IAltitudeAngelService>();
+            if (settings.CheckEnableAltitudeAngel)
+            {
+                await service.SignInIfAuthenticated();
+                return;
+            }
+            if (CustomMessageBox.Show(
```
meee1: also if the user opts out, it nag's please fix this

Utilities/AltitudeAngel/AltitudeAngel.cs
meee1: I didn't notice it was a dynamically referenced function, so I think its ok

rupertbenbrook: Yes, that's correct. MissionPlannerStateAdapter takes a Func<CurrentState> and this is called regularly (twice a second at the moment) to get the current state.

PR #1951, opened 2018-09-30
Exposing sky and ground color values to support upcoming "Night MODE" plugin

EosBandi:
I'm working on a "night mode" plugin, which enables quick switch between different color schemas for night or day usage. To support this, colors used for hud sky and ground are needed to be exposed as public properties. This change does this, keeping the current values as default setting.


EosBandi:
Hi Michael! Any chance that this will make to the next release ? (Not beeing pushy, just try to adjust my release plan with it.)

EosBandi:
Thanks !

PR #1964, opened 2018-10-20
Map: Added Japan Maps

muramura:
I would like to add a map of Japan.
· There are few plains in Japan.
· The height of the terrain is difficult to grasp on Google Maps.
· Make multi-copter fly in the mountains.
· Past images are not hidden like roads.
· You may use a submarine map on the ocean floor.

### DIFF
--- a/ExtLibs/GMap.NET.Core/GMap.NET.MapProviders/Etc/JapanLakeMapProvider.cs
+++ b/ExtLibs/GMap.NET.Core/GMap.NET.MapProviders/Etc/JapanLakeMapProvider.cs
@@ -1,0 +1,95 @@
+﻿namespace GMap.NET.MapProviders
+{
+    using System;
+    using GMap.NET.Projections;
+    using System.Globalization;
+
+    /// <summary>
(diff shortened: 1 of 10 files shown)


meee1:
can I request a change to move the maps libs to MissionPlanner.Maps
and then inject into the instance in program.cs in MP
eg
 // add my custom map providers
            GMap.NET.MapProviders.GMapProviders.List.Add(Maps.WMSProvider.Instance);

muramura:
@meee1 san. thank you for your comment.
I will change.

meee1:
@muramura thanks just trying to minimise changes to the external lib aka gmaps

muramura:
@meee1 san.. I want advice.
I changed it.
Up and down of the line of the map are reversed.
Where should I change and change?
now (NG)
4
3
2
1
before (OK view)
1
2
3
4
ddd

meee1:
@muramura please have a look at Eniro_topo
            ret = string.Format(CultureInfo.InvariantCulture, CustomURL, zoom, pos.X, (1 << zoom) - 1 - pos.Y);
this does a Y axis reverse

PR #1984, opened 2018-11-08
fixed garbled char in message.

YuuyaFujiwara:
I fixed garbled char in message. 
Because of build success.

### DIFF
--- a/ExtLibs/ProjNet/CoordinateSystems/Ellipsoid.cs
+++ b/ExtLibs/ProjNet/CoordinateSystems/Ellipsoid.cs
@@ -142,7 +142,7 @@ namespace ProjNet.CoordinateSystems
 			get
 			{
 				return new Ellipsoid(20926202, 0, 297, true, LinearUnit.ClarkesFoot, "Clarke 1880", "EPSG", 7034, "Clarke 1880", String.Empty,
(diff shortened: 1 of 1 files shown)


ExtLibs/ProjNet/CoordinateSystems/Ellipsoid.cs
```diff
@@ -142,7 +142,7 @@ public static Ellipsoid Clarke1880
 			get
 			{
 				return new Ellipsoid(20926202, 0, 297, true, LinearUnit.ClarkesFoot, "Clarke 1880", "EPSG", 7034, "Clarke 1880", String.Empty,
-					"Clarke gave a and b and also 1/f=293.465 (to 3 decimal places).  1/f derived from a and b = 293.4663077�");
+					"Clarke gave a and b and also 1/f=293.465 (to 3 decimal places).  1/f derived from a and b = 293.4663077.");
```
muramura: I think that the character code "..." is good. I think that the meaning of "." Will change. I think that it is better to use three "." When using "." To represent it.

meee1: I agree, I don't think its a typo at all.
aka https://en.wikipedia.org/wiki/Ellipsis

PR #1986, opened 2018-11-11
Curve item units displayed on graph

TEAM-NAH:
It would be beneficial for the users to be able to know the units of a line as it is displayed on the Mavlink Graph log analyses page. Also fixed a typo that is displayed on the Front end application


TEAM-NAH:
We would love to get feedback on this PR. If the Unit's dictionary should be placed in a resource file or if the current placement of the code is fine. Any tips on how to proceed would be appreciated

meee1:
I like the intent. however i feel it would be better not to have hardcoded entries in the zedgraph library
i made this change https://github.com/ArduPilot/MissionPlanner/commit/1cb7d101d8027fd68567e9bac9e8c744eab7bb7b
that would be more universal.
So it it where me,
modify the zedgraph library to have a unit property per line
then modify the code to pull the attribute and set the unit on the line when the data is graphed.
this makes it usable into the future. ie im thinking using something like this for dataflash logs.

TEAM-NAH:
Thanks for the great feedback. I just wanted to verify with your changes that the units are going to be displayed in some fashion on the Log Analyses Graph, is that correct? I tried pulling your commit and testing the code that you have implemented but was running into exception errors. Thanks for your fast and detailed responses! From not being able to test your code my team is not sure if we should implement our improvements further or if the changes you have committed include what we have envisioned.

PR #2055, opened 2018-12-24
Added QAUTOTUNE to flight modes

tridge:
this mode being missing is really confusing users. I think we need this until
it is in a stable release.

I know this should be automatic when users load firmware, but that isn't working. I suspect it may be because users are using load custom firmware.


rmackay9:
We see this in Copter as well and what I tell users is:

- use the beta MP
- if it doesn't appear press Ctrl-F and push the "paramgen" button

I suspect what needs to be done is MP needs to more proactively go and get the parameters appropriate for the current release.

rmackay9:
@tridge, MP needs to be restarted after the paramgen is run.. and it takes about 2 min and provides no feedback as to when it completes

meee1:
not required.

PR #2075, opened 2019-01-17
OnboardOSD Config tool impl.

SergeyBokhantsev:
This is a config tool for Ardupilot's Onboard OSD.
Now it is only supports online mode (need a connection to a FC). Next step is to implement offline mode (edit .params file).
Conceptually it is designed to generically support any new OSD  items/parameters that could be added later to Ardupilot without need to make a changes to config tool.
Items selecting/positioning by a mouse is supported.

image

image


tridge:
great stuff, thanks!

vierfuffzig:
awesome!

sh83:
Thanks you! Ardupilot OSD should be much more usable with this tool.

tridge:
I've tested this and it works very well
the only issue I hit is that if you enable a screen that wasn't previously enabled then you need to disconnect and reconnect to see it
hmm, one more thing, the "Onboard OSD" option appears even if the board has no OSD parameters. I think it shouldn't show in the side-bar in that case

SergeyBokhantsev:
@tridge thanks,
>>the only issue I hit is that if you enable a screen that wasn't previously enabled then you need to disconnect and reconnect to see it

this because the GUI is dynamically generated basing on the OSD setting the Ardupilot provided to MP. For disabled screens there are no settings so you need to refresh your settings after screen was enabled.
Now there are no explicit [Refresh] button (but you could switch to Parameters list and do refresh from there). Anyway this may be improved by showing a message box after user had enabled a screen: "Do you want to refresh now? Y/N" 

>> the "Onboard OSD" option appears even if the board has no OSD parameters

The MP layout is created before a connection, so maybe @meee1 will advice here, because I didn't found an appropriate hook that could show/hide sidebar items after a connection and parameters receive completion.
As an option could be the MP configuration setting "Show Onboard OSD Configurator"

meee1:
have a look at using IActivate. use this to regenerate the screen

SergeyBokhantsev:
@meee1 yes, I'm using IActivate, no problems here. The question is there a way to dynamically hide or show OnboardOSD item form here, basing on absence or presence of OSD_... settings:
image

knputyi:
Just tested a bit, working like a charm ! Love it !
capture

SergeyBokhantsev:
thanks, @knputyi 
from your screenshot it seems I definitely need to add a scaling option to allow user to make OSD screen 2x smaller. For smaller monitors.

knputyi:
> 
> 
> thanks, @knputyi
> from your screenshot it seems I definitely need to add a scaling option to allow user to make OSD screen 2x smaller. For smaller monitors.

Haha, I need reading glasses, for me, bigger is better !

SergeyBokhantsev:
Ok, then I need to add a scrollbars to the whole tool window. Because currently I don't understand how can you access the OSD Items panel that is below OSD screen.

SergeyBokhantsev:
>>  I accesed the OSD settings in the Parameter list, or tree!
Good workaround, but that is not how I planned the usability :)

knputyi:
I shared the upcoming great news on 2 different FB pages, a LOT of positive feedback in a few hours ! Keep up the good work, waiting for devs to merge it in master.

SergeyBokhantsev:
@meee1 I believe this is near from feature-complete now, please review.
image
image

knputyi:
Ok, nice job again !
One issue : as seeing on first pic., some part from the right side of the window is missing.
Improvement could be made by adding a scroll bar , as I marked with green on second pic.

osd1
osd2

knputyi:
Hi.
Is in Master already ?

SergeyBokhantsev:
Not yet, waiting @meee1 for a review.
Also I will fix your note about scroll bar.

knputyi:
Thanks. It is an outstanding work!
Best regards, Norbert

knputyi:
So, is this project stuck somewhere?

amagr0:
Hello,
I already have FPV+OSD connected to Pixhawk, but I don't see the "Onboard OSD" page in "Config/Tuning". 
Is it necessary to do any activation in order to access to the "Onboard OSD" page ?

Thanks

PR #2096, opened 2019-03-07
Replaced unawaited Task.Delays with Thread.Sleep. 

peancor:
if not awaited Task.Delay returns inmediately and that is not its intended use. Thousands of Tasks/Timer objects could potentially be created with unexpected consecuences.


jjbubka:
This PR solves program hang/freeze while using bluetooth based telemetry, under low signal circumstances.

Issue #2149, opened 2019-05-20
Support for NXP FMUK66 USB VID/PID


jarivanewijk:
NXP Semiconductors has made the RDDRONE-FMUK66 for its HoverGames competition. The board is already supported by PX4 and hopefully will be supported in Ardupilot in the future (I believe there was some early effort some time ago). 

It would be great if support for the following USB VID/PID could be added to the Mission Planner drivers.

USB_VENDOR 0x1FC9
USB_PRODUCT 0x1C
USB_STRING_MANUFACTURER "NXP"
USB_STRING_PRODUCT "RDDRONE-FMUK66"

Thanks. Please let us know if we can contribute in any way, or if more information is needed.

FYI @igalloway, @LeoMustafaNXP, @davids5, @DonLakeFlyer

jarivanewijk:
I found some time to look into this myself. I made a pull request (#2155). Please verify that I made the right changes. Thanks!

meee1:
is the board using the ardupilot bootloader?

davids5:
@meee1 It is using the PX4 bootloader built on top of the NXP_Kinetis_Bootloader_2_0_0.

jarivanewijk:
@meee1 It shouldn't matter whether the board is using the ArduPilot or PX4 bootloader, right?

What we want right now is to get FMUK66 supported in the driver, and to have the board being recognized by Mission Planner. Firmware update through Mission Planner would be nice, but is not really a priority right now.

Could you verify that I made the right changes to achieve this in #2155 or explain what else needs to be done?

PR #2155, opened 2019-06-05
Add new board NXP RDDRONE-FMUK66

jarivanewijk:
Added support for the NXP RDDRONE-FMUK66, more info: #2149 

I'm not very familiar with the MissionPlanner source, so please have a look if I made the right changes and generated and signed the .cat file correctly.

The main goal is to have driver support for now. Detection and firmware update in Mission Planner would be nice, but there is only PX4 support right now (not Ardupilot yet).


DonLakeFlyer:
@meee1 I've lost track of where you are dropping the built version of px4driver.msi. Could you point me to that again. Sorry about that. I'll add some docs for where it comes from to QGC.

magicrub:
Adding support for new hardware is trivial with our chibiOS infrastructure. Has anyone thrown together a hwdef yet?

jarivanewijk:
@magicrub Are you referring to these steps to port ArduPilot onto a new flight controller? http://ardupilot.org/dev/docs/porting.html

It might not be a trivial task because we are using a NXP Kinetis K66, not a STM32F4 or STM32F7. 
I am not sure if there are any current plans within NXP to start porting ArduPilot onto the RDDRONE-FMUK66, but maybe in the future we will pick that up again. I am not in a position to decide.

For our current needs, driver support and detection within Mission Planner should be enough. As far as I'm aware, that's working now.

magicrub:
Ahh yes, my bad. The script architecture is for stm32s.

DonLakeFlyer:
Ok, I did some poking and I think the pre-built driver.msi comes from here: http://firmware.ardupilot.org/Tools/MissionPlanner/driver.msi

If that is correct then:
- The drivers.wxs file was not updated in this pull so the drivers.msi does not include this new driver.
- The name of the install for this is "Drivers" which is not super helpful for users to understand what this is. It shows up as "Drivers" for example in the uninstall list. Can it be changed to something more descriptive. It used to be "PX4 Drivers" which was more useful. Can it be changed to something like "AutoPilot Drivers" or "UAV AutoPilot Drivers" or something like that.

I'm also not sure any more how to detect somehow from the registry whether the latest version of this is installed? It seems like the version is 1.3.7 but that doesn't seem to be updated ever. Is there some other reg keys I can check to determine whether a new version of drivers needs to be installed?

dk7xe:
@DonLakeFlyer @magicrub please give feedback what needs to be done and by who'm to have the driver finally updated.
Thank you, Gerald

meee1:
drivers are available here
http://firmware.ardupilot.org/Tools/MissionPlanner/driver.msi

meee1:
@DonLakeFlyer i think most of your concern have been addressed. the msi uses the same upgrade code as the old package, so should inplace upgrade without any issues.
and is now UAV Driver

DonLakeFlyer:
> i think most of your concern have been addressed.

Thanks so much @meee1. I'll pull in the new version to QGC setups.

dk7xe:
BTW, i have tested the driver on a fresh win10 Installation yesterday by copying the driver folder from git over into the driver folder QGC has generated. FMUK66 is perfectly recognized.

DonLakeFlyer:
@meee1 I put the new drivers into QGC install. But the problem is that I don't see how I can decide whether the driver package needs to be run or not. If I run the package when it already installed then the user gets the "Repair/Remove" variation of setup which is confusing. I don't think I can leave it that way because lots of folks won't know what to do or think something has gone wrong. Is there some way I check whether the drivers are already up to date with the version I have in the QGC installer?

PR #2176, opened 2019-06-27
Change reference to OpenTK.

denis-ivanov:


meee1:
no longer valid,
reason for the direct use was a nuget issue originally

PR #2182, opened 2019-07-16
Fix log->param extraction files to be CSV instead of TSV

magicrub:
Fix log->param extraction files to be CSV (comma) instead of TSV (tab)


magicrub:
@meee1 ping!

meee1:
Need to add localization support to this. Is , vs .   When using tabs it was lass important

magicrub:
Localization shouldn't matter. When you do the usual save Params to file you already use comma so it's just to match that behavior.

PR #2199, opened 2019-08-14
Rover: add waypoint item DO_NAV_DELAY

magicrub:
Sub and Copter has this already, just needs to be added to Rover. It was recently merged into ArduPilot
https://github.com/ArduPilot/ardupilot/pull/12038


rmackay9:
very nice!

PR #2218, opened 2019-09-10
holybro.inf: update Pixhawk6 to Durandal

auturgy:
board has been renamed prior to release.  It would be great to get this in so as not to confuse users.


meee1:
this has been done and signed

PR #2275, opened 2019-12-10
Udp multicast

jfint:
Added a multicast UDP join prompt to UDP communications, and added some values to the rate pull-downs.  I had intended to do these as 2 pull requests, but having some issue doing that with the pull request interface.

### DIFF
--- a/ExtLibs/Comms/CommsUdpSerial.cs
+++ b/ExtLibs/Comms/CommsUdpSerial.cs
@@ -113,6 +113,11 @@ namespace MissionPlanner.Comms
                     "Enter Local port (ensure remote end is already sending)", ref dest)) return;
             Port = dest;
 
+            var multicast_source = "0.0.0.0";
+            if (inputboxreturn.Cancel == OnInputBoxShow("Listen Multicast?",
+                    "Enter Multicast to Subscribe too", ref multicast_source)) return;
+
+
             OnSettings("UDP_port" + ConfigRef, Port, true);
 
             //######################################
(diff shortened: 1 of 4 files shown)


meee1:
current beta MP already support multicast.

use UDPCL, and enter a multicast ip and port. and it will connect.

jfint:
Oh great, will that be in the 1.3.71 release?

jfint:
Also, I just checked the Beta update, it does not seem to join the multicast group.

meee1:
What milticasy address are you using?

jfint:
224.1.0.44:51019

jfint:
I see, you are only allowing 239.x.x.x or 224.0.0.x

meee1:
yep, thats the defaults for sitl ie 239.255.145.50
should really be a range however

jfint:
I'm going to close this request, and possibly try again by opening that up.

jfint:
Thank you

jfint:
@meee1 I wanted to ask here before doing anything else.  The multicast behavior on UDPCL is to send uplink data on the IP and port that is entered, this is generally bad behavior for multicast using the same IP and port for both directions.  The behavior on UDP was to unicast up with Multicast listening, and to uplink on the recievePort+1.  So if Mission planner listened to port 14550, it uplinked on port 14551.  In general this is better behavior.  I think if it multicasted uplink, but on a different port that would at least allow application layer filtering to shelter things from unwanted data.

Do you have any opinions on how you would like it implemented if the IP entered on UDPCL is multicast?

PR #2284, opened 2019-12-25
USB drivers for both single and dual CDC ChibiOS boards

### LINKED ISSUE #2283: new serial drivers break F7 boards
There seems to have been an update to the USB drivers recently which has broken F7 boards very badly. 
For example, if a user has a CUAVv5Nano and they set SERIAL7_PROTOCOl=22 to enable SLCAN on the 2nd USB endpoint then only one USB endpoint appears in windows, and that one endpoint is the SLCAN one. This means the user has no way to fix the issue except by firing up Linux so that can access both endpoints.
The same board shows up correctly as two endpoints in Linux.

tridge:
this uses UDB IDs as follows:

 single endpoint: 0x0483/0x5740
 single endpoint: 0x1209/0x5741
   dual endpoint: 0x1209/0x5740

This is to fix #2283

With this driver installed I see this in MissionPlanner:

image

This PR also removes the stmcdc driver which was on the 0x0483/0x5740 ID. I am not sure if we need to do anything else to get rid of it completely so users get the new driver? Will they need to do a driver clean?
Maybe we need some changes in CleanDrivers.cs too?


meee1:
merged via cherry pick

PR #2290, opened 2020-01-06
Update MainV2.cs to correctly check for windows XP

khancyr:
another solution would be to use RuntimeInformation.IsOSPlatform(OSPlatform.Windows) but that bring another dependency


khancyr:
I don't have Windows XP to test it back, but at least it won't mess up with  docker anymore !

magicrub:
There's gotta be an enum too check against... Right?

khancyr:
Sorry I didn't understand you question ... where do you want an enum ? for 4, 6 and 128 ?

PR #2325, opened 2020-02-06
Do not display waypoints with lat and lon 0

tatsuy:
I often set the lat=0 and lon=0 of waypoint to hold the current location.
but I am not happy that the location of (0, 0) is displayed on the map.

Before
image

After
image

### DIFF
--- a/ExtLibs/Maps/WPOverlay.cs
+++ b/ExtLibs/Maps/WPOverlay.cs
@@ -158,6 +158,10 @@ namespace MissionPlanner.ArduPilot
                         addpolygonmarker((a + 1).ToString(), item.lng, item.lat,
                             item.alt, Color.Green, wpradius);
                     }
+                    else if (command == (ushort)MAVLink.MAV_CMD.WAYPOINT && item.lat == 0 && item.lng == 0)
+                    {
+                        fullpointlist.Add(pointlist[pointlist.Count - 1]);
+                    }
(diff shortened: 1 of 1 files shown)


meee1:
I don't actually like this approach, as it hides a possible user problem. I think the user should be shown an error such as this. Hiding it archives nothing.

tatsuy:
What is a possible user problem?
ArduPilot supports  to hold the current position using WAYPOINT command.
> Lat, Lon - the latitude and longitude targets. If left as zero it will hold the current location.
https://ardupilot.org/copter/docs/mission-command-list.html#waypoint

rmackay9:
In the flight code, the lat/lon of 0,0 is interpreted as stay-at-current-location so I think it makes sense not to show the line back to the west side of Africa.

meee1:
@rmackay9 are you aware if that applys to all codebases?

rmackay9:
@meee1,

It's certainly true of Copter and Plane.  [Here's the code for Plane](https://github.com/ArduPilot/ardupilot/blob/master/ArduPlane/commands.cpp#L31).  I'll check Rover but even if it's not we should make it the same as Copter and Plane.

rmackay9:
@meee1 great, txs!

tatsuy:
@meee1 , @rmackay9 
Thank you for the review and merging.

Issue #2346, opened 2020-03-05
display named value float in Quick status


tridge:
It would be really useful to be able to display NAMED_VALUE_FLOAT values in the MissionPlanner "Quick" display. The NAMED_VALUE_FLOAT is a single float value in mavlink with a string name. If a user could add a "Quick" field and choose the name they want to display then that would allow us to send NAMED_VALUE_FLOAT from on-board lua scripts to display custom values. It would be a massively useful way of extending the ability of lua scripts.
There is one current NAMED_VALUE_FLOAT we send in ArduPilot, the "AS2" value which is the 2nd airspeed sensor. So in SITL you can enable a 2nd airspeed sensor and start getting AS2 NAMED_VALUE_FLOAT for testing.

EosBandi:
OK, should it be automatic, or something that we can define in the config ? I mean just wait for a mavlink message and add the named value to currentstate or we can predefine names in config and look for them when updating currentstate ?

meee1:
im thinking about using a dictionary in currentstate. but need to work though how complicated it will get

EosBandi:
https://github.com/ArduPilot/MissionPlanner/pull/2360 implements support fort NAMED_VALUE_FLOAT display.

IamPete1:
I have just tested this with, https://github.com/ArduPilot/ardupilot/pull/13753 Works great, thanks @EosBandi 

I think this issue can now be closed.

IamPete1:
Having finished messing with named floats I have discovered that it is impossible to remove it from the tuning graph. I had ticked it, now I just get 'customfield0' on the graph but no way to un-tick as its no in the list.

PR #2350, opened 2020-03-09
Extra variables in currenstate for plugin use.

EosBandi:
Currently it is very hard (if not impossible) to add variables to current state from a plugin (cannot display anything in QuickView tab). This PR adds five cs variables (extra1..5) and the possibility to add description to them from settings. (<extra1_desc> etc...) Since these variables are not binding to a datasource they can be directly manipulated from a plugin.


EosBandi:
@meee1 Any thoughts on this ?

magicrub:
LGTM. Would it be worthwhile to make it a double instead of a float?

EosBandi:
The main reason for these variables is to proved means for displaying values on the QuickView tab and on the HUD. I don't see any use case where double precision is needed....

meee1:
ok the reason for the delay, is i think this
https://github.com/ArduPilot/MissionPlanner/issues/2346
could be tied into this as well

EosBandi:
Reimplemented in https://github.com/ArduPilot/MissionPlanner/pull/2360

PR #2352, opened 2020-03-11
DashWare: add trailing commas to csv file

kd0aij:
short packets dont get trailing commas out to the max column number...so those columns get interpreted by Dashware as ZEROS instead of nulls

@Hwurzburg Could you test this fix?

### DIFF
--- a/ExtLibs/Utilities/DashWare.cs
+++ b/ExtLibs/Utilities/DashWare.cs
@@ -35,8 +35,10 @@ namespace MissionPlanner.Utilities
                 using (StreamWriter sr = new StreamWriter(fileout))
                 {
                     // header
+                    var nCols = 0;
                     foreach (var item in colList)
                     {
+                        nCols++;
                         sr.Write(item + ",");
                     }
                     sr.WriteLine();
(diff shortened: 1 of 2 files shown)


Hwurzburg:
Would be happy to..but I have never built MP and dont know how.....convert a log file and dropbox it to me and I will test that...

PS you didnt happen to change the dialog default for what it converts as a default to git rid of CURR and replace with BAT while you were are it, did you?

meee1:
You can get the built version for testing from the appveyor ci link. And click the artifacts section

Hwurzburg:
Works great...while you are tinkering, could you change the CURR default in the dialog to BAT? thanks

kd0aij:
@Hwurzburg Here's a link to a generated CSV file with CURR changed to BAT (manually):
https://drive.google.com/open?id=1bg-cDaFIsx0k3ZFDq9g6P4qSpftkMmzl

@meee1 Is temp.cs a generated file? It's been a long time since I last used Visual Studio.

Hwurzburg:
no need for that....I just created it with BAT instead of CURR..when creating just backspace the CURR and replace with BAT when MP asks what log values to convert...but it would be nice that its changed as a default since that message no longer exists...

kd0aij:
@Hwurzburg That's what I meant :)
I get a bunch of errors when I build locally in VS with the default value change to temp.cs, but it passes CI, so I guess it's OK. As @meee1 pointed out, the build can be downloaded here:
https://ci.appveyor.com/project/meee1/missionplanner/builds/31410078/artifacts

That's pretty nice, I wonder if ArduPilot builds can be downloaded from Travis, Azure or Semaphore?

Hwurzburg:
tested....works great...thanks

PR #2360, opened 2020-03-25
Custom fields for currentstate with support for NAMED_VALUE_FLOAT mav…

EosBandi:
This PR adds 10 fields to CurrentState customfield0-customfield9. They can be used as placeholders for plugins to enable display values in HUD and QuickView.
They can be named by either setting <customfieldx>tag in config or directly manipulating custom_field_names dictionary in CurrentState.

They can be used to display NAMED_VALUE_FLOAT message contents coming from the Flight Controller in two ways.  
1. if we know the name of the value the fields can be named in config.xml as MAV_name this is kind of persistent and guarantee that the same named value will always end up in the same customfield.
2. Dynamically; if a named field message arrives and there is no field named MAV_name then a first available customfield will be allocated to that named value BUT the allocation is not persistent, it not saved to the config. If there are no unnamed customfield then the named field message ignored.

### DIFF
--- a/ExtLibs/ArduPilot/CurrentState.cs
+++ b/ExtLibs/ArduPilot/CurrentState.cs
@@ -178,6 +178,8 @@ namespace MissionPlanner
             ratestatusbackup = 2;
             ratesensorsbackup = 2;
             ratercbackup = 2;
+            //Init dictionary for storing names for customfields
+            custom_field_names = new Dictionary<string, string>();
         }
 
         ~CurrentState()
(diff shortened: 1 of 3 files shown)


thunder-smith:
Hi I've tried to use this to display value in quicview but value doesn't update on quicview. (I'm sending message from arduino)  here's my full question: https://stackoverflow.com/questions/67565496/how-to-properly-display-named-value-float-value-in-quickview

PR #2371, opened 2020-04-19
Add possibility to use own GoogleAPI key

EosBandi:
Due a setup problem (which fixed meanwhile) , if somebody want to use GeoCoding in MP (ZoomTo)  they need their own API key. This line adds a GoogleApiKey parameter to config.xml. If key not found then the original key used.


EosBandi:
It seems that Master does not build at this moment... @meee1 !

meee1:
it does, just not via CI. fixed already though, just not pushed to master

Issue #2026, opened 2018-12-05
Quick Numbers should not be colored


Naterater:
#### Issue details
Having the numbers colored only makes them harder to read. They should default to black or white, and if having color is so important, then the colors should be user-specified, not random as it currently is.

![image](https://user-images.githubusercontent.com/35429159/49549655-01ebdc80-f8af-11e8-8f8e-f33482715cfb.png)

NOTE: Some colors are nearly impossible to read on a bright day with a white background. Especially yellow.

![image](https://user-images.githubusercontent.com/35429159/49549693-25168c00-f8af-11e8-8fd2-230727177d0d.png)

#### Version
All

jeremyfox36:
I'd love to contribute to this project - do you mind if I work on this issue?

PR #2385, opened 2020-05-12
Quick Numbers should not be colored #2026

jeremyfox36:
Sorry I didn't wait for you to give me the go ahead to contribute - I thought it would take me much longer to do this.
I changed ThemeManager.cs (around line 1032):
1) Each QuickView numberColor property is set to the inverse of BGColor.
2) Removes the code that sets each QuickView color individually.


meee1:
this needs to be an option, vs default

PR #2388, opened 2020-05-15
added new RM3100 device type

tridge:
old one is deprecated due to ID change to cope with incorrect scale
factor


tridge:
hold off on merging this for a bit, an issue has come up with RM3100

tridge:
we've now decided not to change the ID, closing

PR #2440, opened 2020-07-31
AltitudeAngel - catch bug crashes with invalid WebBrowser()

magicrub:
I found that my system's WebBrowser() class would return null sometimes when running in debug. I don't know why, but it's at least good to not-crash when it happens. This catches the event and skips the dialog.show which keeps additional problems away


magicrub:
You can either checkout my branch and push a commit to it for the catch handling or merge this and add that afterward on your own branch

magicrub:
closing. This was cherry picked
https://github.com/ArduPilot/MissionPlanner/commit/3c0a1c547ebbc51ee676519a99851adcf0771aca

PR #2453, opened 2020-08-28
HUD message colored by severity

irissam:
When displaying the high-severity statustext to the HUD, color the text depending on its severity.

Use red for severity <= ERROR.
Use yellow for severity <= WARNING.
Use white for less severe than WARNING.


meee1:
ok so my concern with this change is that the last severity state can change the color, but not all instance are set. so the color could be random.   ie set the message without the color = undefined state.

irissam:
> ok so my concern with this change is that the last severity state can change the color, but not all instance are set. so the color could be random. ie set the message without the color = undefined state.

The setter for `messageHigh` always resets the severity to EMERGENCY, so the color should always default to red if severity is not provided. I don't see how the color could reach an undefined state. Does this address your concern, or am I misunderstanding? Thanks for your feedback!

meee1:
looks like you are correct, i miss read the code.

meee1:
i do have other concerns about the entire change to ExtLibs/ArduPilot/Mavlink/MAVLinkInterface.cs

irissam:
I merged master into this branch with Github's merge tool, but I guess that changed all the line endings 🤦‍♂️. I re-performed the merge and updated the branch, so it looks a lot better now. Was this your concern, or did you have other issues with it?

PR #2494, opened 2020-10-24
Option to showe calculated cell value in HUD

EosBandi:
This patch adds a context menu to the HUD, which enables/disables display of calculated battery cell voltage, below the battery voltage. When user enables, it asks for cell count, and uses the value to calculate cell voltage. 
enable/disable status and cell count are stored in the settings.


meee1:
could i request a rebase on this one, any a screenshot of it?

EosBandi:
Here are the screenshots  from two different sized HUD's
hud_cell1
hud_cell2

meee1:
merged via cherrypick. the git history was messy. Have a look at doing a rebase, vs a merge

PR #2500, opened 2020-11-02
mavlink_gps_rtcm_data_t only gets sent once now to reduce transmitted…

StevenHAtRFDesign:
… data.  Geofence upload now runs in a separate thread and a pop up box shows progress.  Now when downloads a geofence from aircraft, checks point index replied is the same as point index requested.  Geofence upload error message now more descriptive.


meee1:
could you split the RTCM change vs the GeoFence change?

StevenHAtRFDesign:
https://github.com/ArduPilot/MissionPlanner/pull/2508
https://github.com/ArduPilot/MissionPlanner/pull/2507

meee1:
merged. thanks

PR #2524, opened 2020-12-22
use "POS" data to draw map

tatsuy:
In case of non-gps navigation, there is no GPS log.

The trajectory of POS is disappeared by selecting item
image

It appeared after my change.
image


meee1:
thanks for the explanation.

Issue #2518, opened 2020-12-16
small SITL launch improvements


tridge:
When using the SITL launch a lot, a few small changes would be useful:
 - a checkbox for skipping the download of the SITL binaries. It can take quite a long time, so if you launch a few times in a row it would be good to be able to skip it
 - finding a location on the map takes a long time. I think there should be a way to enter a latitude/longitude
 - it would be great to be able to select from one of the locations in https://github.com/ArduPilot/ardupilot/blob/master/Tools/autotest/locations.txt

tridge:
This PR should make specifying the location easier:
https://github.com/ArduPilot/ardupilot/pull/16176

tridge:
the above PR is now merged, so specifying --home=LocationName now works in MissionPlanner SITL launch

PR #2565, opened 2021-02-27
Added checkbox "Skip Download"

roberthudson:
As per Tridge's request to skip downloading of SITL binaries

https://github.com/ArduPilot/MissionPlanner/issues/2518

(also adjusted tab order of controls)

### DIFF
--- a/GCSViews/SITL.Designer.cs
+++ b/GCSViews/SITL.Designer.cs
@@ -47,6 +47,7 @@ namespace MissionPlanner.GCSViews
             this.label1 = new System.Windows.Forms.Label();
             this.NUM_heading = new System.Windows.Forms.NumericUpDown();
             this.groupBox4 = new System.Windows.Forms.GroupBox();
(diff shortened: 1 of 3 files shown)


tridge:
I've tested this and it works nicely. It looks like this:
image
if the SITL binary doesn't exist you get a popup saying that it failed to download

roberthudson:
I wasn't sure where to put the checkbox.

Placing it to the right of "Wipe" would mean widening the control. That might have negative consequences on narrower screens.

Feel free to adjust as you see fit.

roberthudson:
Instead of a checkbox, would it be better to simply skip downloading if the binary already exists?

Or is it important to download the latest version? How often do the binaries change?

Sorry for the questions, I'm new here :-)

tridge:
> Or is it important to download the latest version? How often do the binaries change?

they change whenever ArduPilot master changes, which is often several times in a day. We usually do want the latest build, but in the future we might like an option to download one of the stable builds.

Issue #2519, opened 2020-12-16
"do this for all" when loading params


tridge:
When loading a full param file you can quite often get a lot of "marked readonly" and "out of range" dialog boxes. It should be nice to have a "do this for all" checkbox in the dialog to make this is a bit faster

roberthudson:
I took a fresh look at this issue...

I just don't like the current approach we're on. Playing "Whack-a-Mole" with dialog boxes doesn't really make sense.
Imagine this, if in Visual Studio there's a missing reference there might be 100+ compile errors. Do I want to see 100 Messageboxes? Heck no! I don't want to see even **one** Messagebox.
I want to see a **_list_** of errors.

If it were up to me, I'd forget about the "Do for all" checkbox

Here's how I would modify MissionPlanner:
-Click Load Params
-The params all load into an array, like they are now.
-If there are one or more issues, display a new grid

In the grid the user can see all params which are "read only" or "out of range"
They could then Select All/Select None and apply the values or disregard the values. Or rollback the entire load

Besides, the current dialog box is ambiguous. "Do you want to continue?"
Continue what? Continue applying the quirky value? Or continue loading the params?

roberthudson:
Something like ....

![image](https://user-images.githubusercontent.com/21204270/109828400-99fddb00-7c0a-11eb-9b47-5e238b6ab64e.png)

tridge:
> I took a fresh look at this issue...

I like this idea a lot. @meee1  ?

meee1:
yes agreed, much cleaner, and seems alot more fit for purpose.

roberthudson:
Ok, I'm getting to know Mission Planner more and more.
So..... I'm rethinking this issue yet again. We're still on the wrong approach.

Preface at @meee1: No disrespect in the following. Mission Planner is just amazing and you've done an incredible job with it. But with system this complex, sometimes small details get overlooked.
I've been asked to look at one particular issue (the Whack-a Mole dialog game) - so that's where I'm focusing.

There's a bit of a logic flaw. Here's why we're getting all those dialog boxes. They popup on the CellValueChanged event.
This event fires for each row when the grid is loaded. What we need is a bool flag "IsLoading" that is set to true when the grid is populated. Then clear the flag.
While IsLoading=true we want to ignore if the value is ReadOnly. Who cares? We only care IF the user tries to CHANGE it.

Here's what I suggest:
[...]

PR #2568, opened 2021-02-28
Suppress repetitive MessageBoxes

roberthudson:
This addresses @tridge 's request to show "do this for all"
#2519
image

The following approach is used:
There are 2 static classes named CustomMessageBox; one in the  MissionPlanner.MsgBox namespace and one in System. Both need adjusting, as one calls the other.

-The System class has a bunch of overrides. I added an optional parameter DoThisForAllText which, when set, will display a checkbox for example "Do this for all current items"
-The MissionPlanner.MsgBox class, which actually renders the messagebox will add the checkbox to the bottom of the form.

If the checkbox is selected, it will store the selection in a static bool property called DoThisForAll
(I understand this may not be the best approach, but it works. I'm open to ideas how to better pass back the checkbox selection)

in ConfigRawParams.cs I added some bools to track whether the dialog should be suppressed or not. If so, the last choice (Yes/No) is saved and used instead of showing the dialog.

### DIFF
--- a/ExtLibs/Controls/CustomMessageBox.cs
+++ b/ExtLibs/Controls/CustomMessageBox.cs
@@ -13,12 +13,19 @@ namespace MissionPlanner.MsgBox
         const int FORM_Y_MARGIN = 10;
         const int FORM_X_MARGIN = 16;
 
-       public delegate void ThemeManager(Control ctl);
+        public delegate void ThemeManager(Control ctl);
 
         public static event ThemeManager ApplyTheme;
 
         static DialogResult _state = DialogResult.None;
 
+        static CheckBox _chkDoThisForAll = new CheckBox();        
+        public static bool DoThisForAll
+        {
+            get { return _chkDoThisForAll.Checked;}
+            set { _chkDoThisForAll.Checked = false; }
+        }
+        
         public static DialogResult Show(string text)
         {
             return Show(text, string.Empty, MessageBoxButtons.OK, MessageBoxIcon.None);
@@ -34,7 +41,7 @@ namespace MissionPlanner.MsgBox
             return Show(text, caption, buttons, MessageBoxIcon.None);
         }
 
-        public static DialogResult Show(string text, string caption, MessageBoxButtons buttons, MessageBoxIcon icon, string YesText = "Yes", string NoText = "No")
+        public static DialogResult Show(string text, string caption, MessageBoxButtons buttons, MessageBoxIcon icon, string YesText = "Yes", string NoText = "No", string DoThisForAllText = null)
         {
             DialogResult answer = DialogResult.Cancel;
 
@@ -63,13 +70,13 @@ namespace MissionPlanner.MsgBox
             else
             {
                 Console.WriteLine("CustomMessageBox thread running " + System.Threading.Thread.CurrentThread.Name);
-                answer =  ShowUI(text, caption, buttons, icon, YesText, NoText);
+                answer =  ShowUI(text, caption, buttons, icon, YesText, NoText, DoThisForAllText);
             }
 
             return answer;
         }
 
-        static DialogResult ShowUI(string text, string caption, MessageBoxButtons buttons, MessageBoxIcon icon, string YesText = "Yes", string NoText = "No")
+        static DialogResult ShowUI(string text, string caption, MessageBoxButtons buttons, MessageBoxIcon icon, string YesText = "Yes", string NoText = "No", string doThisForAllText = null)
         {
             DialogResult answer = DialogResult.Abort;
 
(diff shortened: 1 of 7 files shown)


tridge:
nice! testing now

tridge:
does this need to be enabled on the various dialogs?
I see this:
image

tridge:
here are some example param files for fixed wing that trigger this:
http://uav.tridgell.net/tmp/testparam/
just start sitl for a plane in MP and alternately load those two param files, it gives the dialog for out of range on the RLL2SRV params

roberthudson:
Sorry! I somehow missed adding the custom dialog to the Full Parameter Tree. This is fixed in the March 1 commit.
While I was at it, I also added the ReadOnly check to Full Parameter Tree - it was missing.

tridge:
I've tested this and it works really well now, thanks!

tridge:
just found a small issue. After I finish loading the params when I have selected the "for all" checkbox and I then go to load a different parameter file I don't get any warning dialogs at all. I think the correct behaviour would be that the state would clear after the end of the current operation

roberthudson:
Several fixes in this commit....

1.  if there are several missing params, the messagebox wraps and limits to (default of)  25
image

It used to be this ridiculous mile-wide beast
image


2. DoThisForAll flag resets when a new config file is loaded

3. Fixes a null error when the param grid contains an empty cell. In some cases row.Cells[1].Value could be null. 
changed
`if (row.Cells[1].Value.ToString() != value)`
to
`if (row.Cells[1].Value != null && row.Cells[1].Value.ToString() != value)`

(Thanks for spotting that tridge, I'm surprised this wasn't detected long ago)

4. Added a few try/catches

roberthudson:
> yeah, having a static seems wrong here, and just prone to issues.

True, I realize it's not ideal. But what's the alternative?

Yeah it's a static, but it's just a simple boolean flag. There isn't really a thread-safety issue. And the flag gets cleared each time the messagebox is displayed.

My first idea was adding some more enums to the DialogResut (adding OKToAll, YesToAll, NoToAll)

`public enum DialogResult
        {
            None,
            OK,
            Cancel,
            Abort,
            Retry,
            Ignore,
            Yes,
            No,
            OKToAll,
            YesToAll,
            NoToAll
        }`

But that won't work. 
The MissionPlanner.MsgBox.CustomMessageBox returns a System enum which of course is sealed.

What do you suggest? How else is the calling function going to know the user checked "Do this for all"?

tridge:
> What do you suggest? How else is the calling function going to know the user checked "Do this for all"?

I'm not a C# dev, so this may not work, but perhaps CustomMessageBox could be subclassed to CustomMessageBoxCheckAll and the functionality and static be only in the subclass. Then we could use the subclass just in the places it makes sense, such as the param interface, without risking unexpected impact on other use cases.
There seems to be over 900 uses of CustomMessageBox, and I think it would be easier to opt-in the cases that matter

Common.cs
```diff
@@ -237,5 +238,19 @@ static void chk_CheckStateChanged(object sender, EventArgs e)
...
+
+        public static string getMissingParamsMessage(int missed, System.Collections.Generic.List<string> missing, int maxItems = 25)
+        {
+            string list = "Missing " + missed + " params\n";
+
+            foreach (var item in missing.Take(maxItems))
+            {
+                list += item + "\n";
+            }
+            if (missing.Count > 25)
```
tridge, line 250: i think the 25 here should be maxItems

Common.cs
```diff
@@ -237,5 +238,19 @@ static void chk_CheckStateChanged(object sender, EventArgs e)
...
+                list += "(and " + (missing.Count - 25) + " more).";
```
tridge, line 251: and here

meee1:
i do agree with tridge, it should be opt-in. as the dialog itself is used everywhere, and unrelated to most likely 850 of those 900 usages

tridge (changes requested):

Common.cs
roberthudson: Yeah right. I added that param very last minute - it should be maxItems

Common.cs
roberthudson: See https://github.com/ArduPilot/MissionPlanner/issues/2519

Common.cs
roberthudson: See https://github.com/ArduPilot/MissionPlanner/issues/2519

roberthudson:
Ok, I'm getting to know Mission Planner more and more.
So.....  I'm rethinking this issue yet again. We're _**still**_ on the wrong approach.

Preface at @meee1: No disrespect in the following. Mission Planner is just amazing and you've done an incredible job with it. But with system this complex, sometimes small details get overlooked. 
I've been asked to look at one particular issue (the Whack-a Mole dialog game) - so that's where I'm focusing. 

There's a bit of a logic flaw. Here's why we're getting all those dialog boxes. They popup on the CellValueChanged event. 
This event fires for each row when the grid is loaded. What we need is a bool flag "IsLoading" that is set to true when the grid is populated. Then clear the flag.
While IsLoading=true we want to ignore if the value is ReadOnly. Who cares? We only care IF the user tries to CHANGE it.

Here's what I suggest:
-set an IsLoading flag and suppress the MessageBox while the grid is populating. Then clear the flag.

-~~then, if the user tries to change a ReadOnly cell we show the dialog.~~ Lock the cell so they can't change it. Mark the cell in orange or something. 

-we can scrap the Do This For All checkbox. It's not needed

-for the Out Of Range cells, mark them in some other color as a visual cue. Maybe yellow

-after populating the grid display a dialog "n values are out of range. They have been highlighted yellow"

PR #2571, opened 2021-03-09
Update ConfigRawParamsTree

EosBandi:
Handle comma/period decimal separator issue. (see same update for ConfigRawParams as well in the next PR)


EosBandi:
@meee1 Did you had time to take a look at this ? I think it is quite serious issue in some countries.

PR #2583, opened 2021-03-30
ConfigurationView: added two parameters in rover basic tuning view

tajisoft:
I think this is very useful for users who tune Rover.


rmackay9:
We don't actually use the ATC_SPEED_FF though actually.  The feed forward comes from the CRUISE_SPEED and CRUISE_THROTTLE.  _SMAX might be useful, I've never used it but it could be.

We should probably modify the parameter description for ATC_SPEED_FF to tell people it is unused and they should set CRUISE_SPEED or CRUISE_THROTTLE instead.

tajisoft:
@rmackay9 Thank you for comment. It seems that my understanding was a little wrong. This PR doesn't make much sense, so I think it will be closed.

PR #2595, opened 2021-04-27
Global: delete unused variables

muramura:
I found a warning about unused variables.
I know that many warnings can obscure critical warnings.
I want to reduce the number of warnings by one or more.


rmackay9:
I agree we should try and reduce the number of warnings.  300+ means there is no chance anyone will actually look at them.

It looks like this PR has a lot of other changes included though.  Perhaps they are white space changes?

PR #2599, opened 2021-05-02
Global: Responding to warning cs0612

muramura:
I want to change the warning CS0612.
I want to reduce the warning.
I want to clarify the real warnings from the many warnings.


meee1:
i think there are issues with this one, there are multiple things tag obsolete that i dont think should be

muramura:
So displaying a warning would encourage people to fix it, which is what we're doing now.
I understood that this in itself is really a warning to change.

PR #2603, opened 2021-05-03
Global: Responding to warning cs0162

muramura:
I have changed the build warning CS0162.
I am unsure of the real warning, to many warnings.
I want to reduce the warnings.

### DIFF
--- a/ExtLibs/MissionPlanner.Drawing/Graphics.cs
+++ b/ExtLibs/MissionPlanner.Drawing/Graphics.cs
@@ -1606,7 +1606,7 @@ GRBackendRenderTargetDesc backendRenderTargetDescription = new GRBackendRenderTa
 
             //get client rect
             //hwnd to hdc
-            return HwndToGraphics(windowHandle);
+            //return HwndToGraphics(windowHandle);
         }
 
         public void ReleaseHdc()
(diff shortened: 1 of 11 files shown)


ExtLibs/MissionPlanner.Drawing/Graphics.cs
```diff
@@ -1606,7 +1606,7 @@ public static Graphics FromHwnd(IntPtr windowHandle)
 
             //get client rect
             //hwnd to hdc
-            return HwndToGraphics(windowHandle);
+            //return HwndToGraphics(windowHandle);
```
meee1, line 1609: i question this?

ExtLibs/UAVCAN/Extension.cs
```diff
@@ -39,19 +39,14 @@ public static object GetValue(this uavcan.uavcan_protocol_param_Value value)
             {
                 case uavcan.uavcan_protocol_param_Value_type_t.UAVCAN_PROTOCOL_PARAM_VALUE_TYPE_BOOLEAN_VALUE:
                     return value.union.boolean_value;
-                    break;
```
meee1, line 42: the breaks should be there, as i dont want fallthough

ExtLibs/MissionPlanner.Drawing/Graphics.cs
muramura: This method has two RETURNs in the sequence.
The first RETURN will be executed and this RETURN will not be executed permanently.
I have commented out this RETURN handling as I think it will be useful in the future.

ExtLibs/UAVCAN/Extension.cs
muramura: MP is a coding style that leaves the CASE-BREAK structure intact and describes the process.
I thought this was unnecessary, because I have never experienced that the RETURN process is not executed.

Issue #2613, opened 2021-05-10
"Change Speed" button's default is too high, remove set home alt for Copter, Rover


rmackay9:
A user was struggling to use the "Change Speed" button so I noticed that the default value is insanely high (100m/s).  We should really reduce this to a more sensible number like 10 (m/s).

As a side note, I don't think it's not a good idea to share the box with the "Set Home Alt" in fact, I think we should remove this box for all vehicles except perhaps Plane that might want to keep it.

![image](https://user-images.githubusercontent.com/1498098/117591978-b3bf1f80-b171-11eb-8e52-9ffabc041eee.png)

jagrandle:
This sounds like a great idea.

Planes and copter often have quite different speeds. Is it proposed to have just one default speed appear here for different types of aircraft ?

Would it be possible to have it default to the Trim or Cruise speed of a fixed wing and maybe the Mission or Loiter speed for multirotor, or similar ? This would potentially reduce the opportunity for drastic mistakes.

Cheers

PR #2615, opened 2021-05-15
GCSViews: Set default value to 10 in response to issue 2613

muramura:
I responded to issue #2613.
I changed the default value from 100 to 10.

Screenshot from 2021-05-15 14-06-50


jagrandle:
Is the default of 10 only for multirotor or also fixed wing ?
If it is for fixed wing then it may be more dangerous than leaving at 100.
Surely there should be some discussion about such changes before they are made for default for everybody ?

PR #2623, opened 2021-05-25
Params:   cache FTP downloaded params to ftpparam.pck after mavftping them....

davidbuzz:
if file exists on-disk, read params from that file in preference to doing mavftp  after that. 

Warning, this file is currently *never* removed by the system.  

I'm hoping that each time MP does a mavftp of the params it will overwrite it, but i haven't really tested that.


meee1:
after having a trial of caching. its going to be a opt in feature. as there are many caveats that create problems. please check current master for the current implementation

PR #2649, opened 2021-06-17
Updated Config -> Extended Tuning to include new filter values

AndresUsandi:
In the last few releases, several new filter values have been added, including notch and harmonic filters and FLT* filters. I have updated the UI to include fields for all those filters so that they don't need to be modified from the the parameter list or tree views.


AndresUsandi:
This is what the new UI looks like
FiltersUI

rmackay9:
fancy!  nice!

PR #2654, opened 2021-06-26
GCSViews: If the delay of WAYPOINT is set to less than 0, set it to 0

muramura:
I set it to 0 if the WAYPOINT delay is set to a value less than 0.
I set the PR to return an error response if the WAYPOINT delay is set to a value less than 0 on the ArduPilot side.
I hoped that the ArduPilot members would check the input value on the GCS side.
I was unable to set a value less than 0 for the WAYPOINT delay in APM PLANNER2.
I decided to change the MP side.

https://github.com/ArduPilot/ardupilot/pull/17704

### DIFF
--- a/GCSViews/FlightPlanner.cs
+++ b/GCSViews/FlightPlanner.cs
@@ -2979,7 +2979,16 @@ namespace MissionPlanner.GCSViews
                             false);
                 }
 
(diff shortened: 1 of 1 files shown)


meee1 (changes requested):
please remove the mono submodule changes

muramura:
I committed again, to undo the mono commit.

muramura:
The reworked PR is up and closed.

PR #2665, opened 2021-07-28
FlightPlanner: Clear selected marker on ContextMenu open

EosBandi:
It fixes, shen a context menu is opened over a marker, it remains selected and after closing the menu (or called dialog) next click on the map moves the marker.


meee1:
ive not tested this, but do thing that are based off the context menu still work? like "delete"

EosBandi:
You are right, it breaks that functionality. I will reevaluate how to fix it. Closing it now.

PR #2692, opened 2021-09-28
LogSort: Add sorting by BRD_SER_NUM

TunaLobster:
I'm not sure how well this works because I'm not able to get a VS environment set up.

The intent is to also sort log files based on BRD_SER_NUM. There are some uses for that parameter to identify what autopilot was used in which plane. Changes or other ideas are welcome. Maybe SYSID-SER_NUM format?

### DIFF
--- a/ExtLibs/Utilities/LogSort.cs
+++ b/ExtLibs/Utilities/LogSort.cs
@@ -68,6 +68,7 @@ namespace MissionPlanner.Log
                     bool issitl = false;
                     var sysid = 0;
                     var compid = 0;
+                    var brdsernum = 0;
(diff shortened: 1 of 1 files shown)


meee1:
so the concept is good, the issue is that this is only changing the sitl logs atm, ie to really be usefull needs to include the others as well. i would also add a check for  "0"  to exclude it

TunaLobster:
@meee1 I'm struggling a little bit to get this working in practice. Comments on what I am doing with the code would be helpful. Trying to learn a bit as I go.

magicrub:
This is a great idea!

PR #2716, opened 2021-11-03
Grid: Use PlannedHomeLocation altitude

EosBandi:
Camera footprints are not displayed if we don't have valid home altitude. This patch change startalt to PlannedHomeLocation.Alt.
Now footprints are shown again :)


meee1:
@EosBandi thanks you

Issue #2717, opened 2021-11-05
failed to load fw when running px4 bootloader


tridge:
when a user buys a board that came with px4 and wants to use ArduPilot they will want to use MissionPlanner to load the firmware. That doesn't work at the moment, I suspect because of the recent extflash support. I applied a fix for this to Tools/scripts/uploader.py that I think needs to be applied in MissionPlanner as well.
This was the fix needed for uploader.py:
https://github.com/ArduPilot/ardupilot/commit/dfd26868780084148bdb8e833fd5f33a05175e8c#diff-c16aa82ddeb6577c91570778032de965e3c728dfd7dc668abb2bf7a03368702f
There are two key changes:
 - check for ext flash size before the board_type, board_rev and fw_maxsize checks
 - resync after a failed extflash check

Here is what happens with a new out of the box Pixhawk5x.
![image](https://user-images.githubusercontent.com/831867/140570017-257d4c0e-92a9-4063-a791-5f6ae6cc49da.png)

MissionPlanner log:
![image](https://user-images.githubusercontent.com/831867/140572294-30421870-9b4d-45be-a27a-6060cd3ba538.png)

looking at the Uploader.cs, I don't think MissionPlanner is trying to get the ext flash size yet, so something else is wrong

tridge:
more debug, I added some print() calls to show leading byte (command code) in each send[]
![image](https://user-images.githubusercontent.com/831867/140579125-5fcb24d0-b312-4aba-b4cb-99d5e67c30d2.png)
the sequence of commands looks fine .... don't know why it doesn't accept it

meee1:
Do we know what bootloader code version is on it?

tridge:
> Do we know what bootloader code version is on it?

it says blrev 5, but that doesn't say much

tridge:
i've tried lots of things to debug with no luck. You can see my various hacks here:
https://github.com/tridge/MissionPlanner/commits/pr-upload-px4-bl
(I use CI to do the MissionPlanner build as I've never managed to get builds working on my win10 box)
another data point: fw load of ardupilot works fine with qgc

PR #2719, opened 2021-11-07
Uploader: re-fetch BOARD_REV before erase

tridge:
this avoids a bug in older versions of the px4 bootloader
fixes issue #2717


tridge:
already fixed in master. Thanks @meee1 !

PR #2726, opened 2021-11-17
GCSVews: Configuration: Compass2: support new compass param names

IamPete1:
My first go a `C#`, so apologies if this is a bit of a mess. 

This adds support for the new compass params names proposed in https://github.com/ArduPilot/ardupilot/pull/19123 and allows both new and old to work as before.

### DIFF
--- a/GCSViews/ConfigurationView/ConfigHWCompass2.cs
+++ b/GCSViews/ConfigurationView/ConfigHWCompass2.cs
@@ -30,17 +30,17 @@ namespace MissionPlanner.GCSViews.ConfigurationView
             public CompassDeviceInfo(int index, string ParamName, uint id) : base(index, ParamName, id)
             {
                 //set initial state
-                var id1 = MainV2.comPort.MAV.param["COMPASS_DEV_ID"];
-                var id2 = MainV2.comPort.MAV.param["COMPASS_DEV_ID2"];
-                var id3 = MainV2.comPort.MAV.param["COMPASS_DEV_ID3"];
+                var id1 = MainV2.comPort.MAV.param[new[] { "COMPASS_DEV_ID", "COMPASS1_DEV_ID"}];
+                var id2 = MainV2.comPort.MAV.param[new[] { "COMPASS_DEV_ID2", "COMPASS2_DEV_ID"}];
+                var id3 = MainV2.comPort.MAV.param[new[] { "COMPASS_DEV_ID3", "COMPASS3_DEV_ID"}];
 
-                var idO1 = MainV2.comPort.MAV.param["COMPASS_ORIENT"];
-                var idO2 = MainV2.comPort.MAV.param["COMPASS_ORIENT2"];
-                var idO3 = MainV2.comPort.MAV.param["COMPASS_ORIENT3"];
+                var idO1 = MainV2.comPort.MAV.param[new[] { "COMPASS_ORIENT", "COMPASS1_ORIENT"}];
+                var idO2 = MainV2.comPort.MAV.param[new[] { "COMPASS_ORIENT2", "COMPASS2_ORIENT"}];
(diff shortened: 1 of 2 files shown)


GCSViews/ConfigurationView/ConfigHWCompass2.cs
```diff
@@ -27,20 +27,20 @@ public class CompassDeviceInfo : DeviceInfo
             private string _orient;
             private bool _external;
 
-            public CompassDeviceInfo(int index, string ParamName, uint id) : base(index, ParamName, id)
+            public CompassDeviceInfo(int index, string ParamName, uint id, ConfigHWCompass2 compass2) : base(index, ParamName, id)
```
meee1, line 30: i know its your first attempt, but ill just point out stuff

instead of passing the instance, i would have made the function static, so it can be used throughout without needing the instance version

GCSViews/ConfigurationView/ConfigHWCompass2.cs
```diff
@@ -27,20 +27,20 @@ public class CompassDeviceInfo : DeviceInfo
...
             private bool _external;
 
-            public CompassDeviceInfo(int index, string ParamName, uint id) : base(index, ParamName, id)
+            public CompassDeviceInfo(int index, string ParamName, uint id, ConfigHWCompass2 compass2) : base(index, ParamName, id)
             {
                 //set initial state
-                var id1 = MainV2.comPort.MAV.param["COMPASS_DEV_ID"];
-                var id2 = MainV2.comPort.MAV.param["COMPASS_DEV_ID2"];
-                var id3 = MainV2.comPort.MAV.param["COMPASS_DEV_ID3"];
+                var id1 = MainV2.comPort.MAV.param[compass2.get_compass_param_name("DEV_ID", 1)];
```
meee1, line 33: static version would be ConfigHWCompass2.get_compass_param_name("DEV_ID", 1)

GCSViews/ConfigurationView/ConfigHWCompass2.cs
```diff
@@ -121,15 +124,15 @@ public void Activate()
             mavlinkComboBoxfitness.setup(ParameterMetaDataRepository.GetParameterOptionsInt("COMPASS_CAL_FIT",
                 MainV2.comPort.MAV.cs.firmware.ToString()), "COMPASS_CAL_FIT", MainV2.comPort.MAV.param);
 
-            mavlinkCheckBoxUseCompass1.setup(1, 0, "COMPASS_USE", MainV2.comPort.MAV.param);
-            mavlinkCheckBoxUseCompass2.setup(1, 0, "COMPASS_USE2", MainV2.comPort.MAV.param);
-            mavlinkCheckBoxUseCompass3.setup(1, 0, "COMPASS_USE3", MainV2.comPort.MAV.param);
+            mavlinkCheckBoxUseCompass1.setup(1, 0, get_compass_param_name("USE", 1), MainV2.comPort.MAV.param);
```
meee1, line 127: there is 2 options here, use what you have done, or pass in both text versions, as the param name field can take an array (havn't checked this, but could be added, as param renames are common these days)

GCSViews/ConfigurationView/ConfigHWCompass2.cs
```diff
@@ -529,5 +532,43 @@ private async void but_missing_ClickAsync(object sender, EventArgs e)
 
             await UpdateFirst3();
         }
+
+        private bool param_format_initialized;
+        private bool param_format_new;
+        private string get_compass_param_name(string param, int index)
```
meee1, line 538: make it static, not sure if the private bools are really required

IamPete1:
@meee1 your right the array of names method is way better, I have changed to that. I had to add a new method to the param list, but it seem to work as expected.

meee1:
nice, i like the approch

PR #2732, opened 2021-11-21
Fix application local data path in *nix based OS

silentjet:
In Linux/Unix like operating systems 'MyDocuments' dir does not exists.
However, the mono framework for compatibility purposes points it to the root of the user directory instead(/home/username). That is incorrect from the point of view of the Linux/Unit standards (see [1] for more detail). The Local Application
Data directory should be used instead(/home/username/.local/share). Fix that for operating systems where platform ID is Unix.

 * [1] https://www.freedesktop.org/wiki/Specifications/basedir-spec/

Please be aware for existing users it might be required to execute a command
```bash
mv "~/Mission Planner" ~/.local/share/
```

### DIFF
--- a/ExtLibs/Utilities/Settings.cs
+++ b/ExtLibs/Utilities/Settings.cs
@@ -289,6 +289,8 @@ namespace MissionPlanner.Utilities
             return (t != null);
         }
 
+        public static bool isUnix = Environment.OSVersion.Platform == PlatformID.Unix;
+
         /// <summary>
         /// Shared data directory
         /// </summary>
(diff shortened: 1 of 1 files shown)


meee1:
ive got no issues with the change, its more around a migration strategy I'm concerned with

silentjet:
> ive got no issues with the change, its more around a migration strategy I'm concerned with

Would you like me to add a directory migration related code? If so, what would be a reasonable place for it in your opinion?

meee1:
ok, so i think the cleanest path is to check if the old directory exists, if it doesnt use your new path. this way it doesnt just break existing users. but all new users will use the new path

silentjet:
> ok, so i think the cleanest path is to check if the old directory exists, if it doesnt use your new path. this way it doesnt just break existing users. but all new users will use the new path

Ok, reworked following your recommendation...

PR #2753, opened 2021-12-20
Speed up ConfigRawParams and ConfigRawParamsTree reload time

EosBandi:
A new Planner setting checkbox called "Running on a slow computer" controls how the ConfigRawParams and ConfigRawParamsTree pages are reloaded. If enabled, it keeps the metadata info and refreshes only the values and the control. For safety, there is a button to refresh the whole table (without reloading parameters from the vehicle).

Time improvements: Measured on an Intel Baytrail T Z3735F Quad Core stick computer with 2GB RAM 
                                              original        Running on Slow Computer
ConfigRawParams reload        13sec                             2.4 sec
ConfigRawParamsTree reload  20sec                             3.5 sec

Note: Initial opening times of these pages are not improved; that needs a significant rewrite of the ParameterMetadataRpository, which is not in the scope yet.

### DIFF
--- a/GCSViews/ConfigurationView/ConfigPlanner.Designer.cs
+++ b/GCSViews/ConfigurationView/ConfigPlanner.Designer.cs
@@ -105,6 +105,7 @@
             this.num_gcsid = new System.Windows.Forms.NumericUpDown();
             this.label7 = new System.Windows.Forms.Label();
             this.CHK_params_bg = new System.Windows.Forms.CheckBox();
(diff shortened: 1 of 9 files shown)


meee1:
this doesn't violate any cross thread gui calls does it?  ie the paralell.for's etc would

EosBandi:
Paralell fors was there, I simply made the rowlist static. None of the two controls are used elsewhere and it just tests OK. So it did not break anything during testing.

Issue #2749, opened 2021-12-13
OSD Color not changeable


GrahamDyer:
#### Issue details
On previous versions of Mission Planner, the color of the OSD text could be changed in the Config/Planner tab, but this no longer appears to be working? Changing the color doesn’t seem to stick/save on the OSD.

#### Version
MP v1.3.75

#### Platform
[ X] All
[  ] AntennaTracker
[  ] Copter
[  ] Plane
[  ] Rover
[  ] Sub

#### Airframe type
Fixed Wing

#### Hardware type
AUAV-X2

#### Logs
N/A

![MPosd_col3](https://user-images.githubusercontent.com/3287001/145820135-93693cb5-56de-4c3f-86db-da4bd052a978.jpg)

![MPosd_col2](https://user-images.githubusercontent.com/3287001/145819455-ea828f5f-74f8-4a7f-b18b-687084aa86fa.jpg)

PR #2759, opened 2021-12-31
Fix for ISSUE #2749, OSD Color not changeable

jawad111:
Fix for ISSUE #2749.
Under Configuration View, The OSD Color parameter changes are now reflected on the OSD. 

after
after2


GrahamDyer:
Any reason why this is not working in v1.3.80

PR #2760, opened 2022-01-01
FlightPlanner: display unknown mission items as UNKNOWN(nnn)

tridge:
this is mostly working, but still has some glitches
The UNKNOWN(nnn) is correctly displayed on the 2nd instance of an unknown ID, but not the first. If you refresh waypoints then all are correctly displayed. When the UNKNOWN(nnn) is first substituted for some reason it doesn't show in the list
image

Update: it now shows all UNKNOWN(xxx) on the first view correctly, so only remaining issue seems to be that when you use "Add Below" to add an UNKNOWN and it asks for the ID, that the WP shows as "UNKNOWN" not UNKNOWN(xxx)


tridge:
closed in favor of #2761

PR #2761, opened 2022-01-02
CTRL-F screen, add editor to add/delete mavlink commands to Planner grid dropdown

EosBandi:
This one is for the Devs. A quick editor enables users to add mavlink commands to the dropdown list of the Planner grid. You can add commands that are already defined in MAV_CMD enum and new ones as well. Also, you can edit the param names that are displayed In the header.
The commands are stored in config.xml.
 Adding a command here does not mean that the autopilot will accept or execute it. So the user has to know what is he/her doing.

### DIFF
--- a/Controls/MavCommandSelection.Designer.cs
+++ b/Controls/MavCommandSelection.Designer.cs
@@ -1,0 +1,196 @@
+﻿
+namespace MissionPlanner.Controls
+{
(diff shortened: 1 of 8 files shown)


tridge:
tested and works very nicely!

tridge:
image
image

Issue #2707, opened 2021-10-26
Plan: Survey (Grid) appears twice in Auto WP right-mouse-button-menu


rmackay9:
On the Plan screen, if you right-mouse-button-click and open the "Auto WP" menu you'll see that "Simple Grid" appears twice.
![image](https://user-images.githubusercontent.com/1498098/138866388-86f909a7-679b-49cd-8d73-1260bed823fa.png)

Also I think that some consolidation of the three similar options, SimpleGrid, Survey (Grid) and Survey (Gridv2) would be good or at least all the items in this menu should be placed in alphabetical order.

Also "Create Circle Survey" should be renamed to "Survey (Circle)" to be consistent with the naming of other items on the list.

rmackay9:
This has been fixed it seems so closing

PR #2764, opened 2022-01-04
Extract GRID plugin from MissionPlanner project

EosBandi:
Put it in its own project in the plugins project folder, alongside with the other plugins (FaceMap et all...) 
Also removed fixed Toolstrip menuitem from FlightPlanner. Now GRID can be disabled as any other plugins.
(It eventually fixes https://github.com/ArduPilot/MissionPlanner/issues/2707 ☺ )


meee1:
it actualy use to be a plugin, then i move it back into MP, as i wanted to use the grid internal functions from within the main code without replicateing code. randys issue is that his install is OLD, and he still have the plugin and the main code version being init'ed

EosBandi:
Opps, my bad. Then I'll close this PR.

PR #2779, opened 2022-01-26
display a warning for bad airspeed health

tridge:
this allows users to see when airspeed is being rejected by:
https://github.com/ArduPilot/ardupilot/pull/19883


meee1:
only issue is the missing resource entry, normaly the *.designer.cs files are autogenerated from the *.resx files
https://github.com/ArduPilot/MissionPlanner/blob/master/ExtLibs/Strings/Strings.resx

tridge:
> only issue is the missing resource entry, normaly the *.designer.cs files are autogenerated from the *.resx files

thanks Michael, I've added the change to the resx file. I'm not sure how to trigger the auto-generation, so I left the manual edit of the designer.cs file - is that ok?

meee1:
yeah, you need to use Visual studio for the resgen process to be run. ive merged anyway

PR #2826, opened 2022-03-31
Fix bugs in external flash flashing

andyp1per:
Allow the flashing of external flash images to boards with external flash

Tested on an SPRacingH7

### DIFF
--- a/ExtLibs/px4uploader/Uploader.cs
+++ b/ExtLibs/px4uploader/Uploader.cs
@@ -464,7 +464,7 @@ namespace px4uploader
             while(port.BytesToRead == 0)
             {
                 if (DateTime.Now > deadline)
-                    throw new TimeoutException("timeout waiting for responce");
+                    throw new TimeoutException("timeout waiting for response");
                 Thread.Yield();
             }
 
@@ -500,6 +500,8 @@ namespace px4uploader
             byte c = __recv()[0];
             if (c != (byte)Code.INSYNC)
                 return false;
+
+            c = __recv()[0];
             if (c != (byte)Code.OK)
                 return false;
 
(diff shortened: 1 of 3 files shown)


ExtLibs/px4uploader/Uploader.cs
```diff
@@ -500,6 +500,8 @@ public bool __trySync()
             byte c = __recv()[0];
             if (c != (byte)Code.INSYNC)
                 return false;
+
+            c = __recv()[0];
```
meee1, line 504: Not sure about this change, why the extra read?

ExtLibs/px4uploader/Uploader.cs
```diff
@@ -742,25 +794,52 @@ public void __verify_extf(Firmware fw)
...
+
             this.port.ReadTimeout = 1000; // 1 sec
 
-            int expect_crc = fw.crc(self.fw_maxsize);
-            __send(new byte[] {(byte)Code.GET_CRC,
-				(byte) Code.EOC});
-            int report_crc = __recv_int();
-            self.__getSync();
+            if (self.fw_maxsize > 0 && false)
+            {
```
meee1, line 807: && false?

ExtLibs/px4uploader/Uploader.cs
andyp1per: Look at the code for sync() - it is absolutely required. This function clealy never worked.

ExtLibs/px4uploader/Uploader.cs
andyp1per: oops, will remove

ExtLibs/px4uploader/Uploader.cs
meee1: indeed, a 7 year old bug.....

andyp1per: Hah, yes - I only found it by comparing with the python uploader

ExtLibs/px4uploader/Uploader.cs
andyp1per: Fixed

andyp1per:
Thanks @meee1 !

PR #2840, opened 2022-04-16
GCSViews: enable intial setup wizard for Quadplane

IamPete1:
Does what it says on the tin.


Hwurzburg:
could we add FF also, wont hurt for copter? ,please please

IamPete1:
@Hwurzburg This initial setup tool is just a guess at filter params from prop size. I'm not sure how we can get FF from that?

Hwurzburg:
sorry thought this was the tuning box....it needs ff added

PR #2843, opened 2022-04-20
GCSViews: ConfigFirmwareManifest: only show one sort of copter

IamPete1:
This cuts down the install firmware page to only show a single copter:

image

The existing buttons all do the same thing anyway. This should result in less confusion as users may think that they need a frame specific firmware. 

The legacy firmware install page is unchanged. 

@rmackay9 any thoughts?


meee1:
the funny thing is ive had a request for the oposite, this was to do with plane/quadplane though.  as soon as you add an icon, that doesnt reflect what the user is trying to do it gets messy.

PR #2847, opened 2022-04-26
ArduPilot: State fix Battery Voltage Reporting for Batts >65.5V

hendjoshsr71:
Request from ArduPilot Partner

This allows the HUD to report voltages greater than 65V. This limitation comes from the mavlink fields only being uint16s.

In ArduPilot we fill the cell fields to overcome this deficiency even if the monitor type doesn't report cell voltages.

I'm not sure how to make this compatible with systems not reporting the `battery_status` message. But I'm not sure  how old that message is or how important compatibility is with systems not reporting battery_status?

I think the battery2 message isn't used anymore as well?

image

### DIFF
--- a/ExtLibs/ArduPilot/CurrentState.cs
+++ b/ExtLibs/ArduPilot/CurrentState.cs
@@ -2537,7 +2537,6 @@ namespace MissionPlanner
 
                             load = sysstatus.load / 10.0f;
 
-                            battery_voltage = sysstatus.voltage_battery / 1000.0f;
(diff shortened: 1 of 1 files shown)


meee1:
are you aware of what version/s this came in at?

hendjoshsr71:
> are you aware of what version/s this came in at?

It looks like this was merged in ArduPilot 4 years ago here https://github.com/ArduPilot/ardupilot/pull/8999/files

And the first stable was Copter 4.0.0

PR #2850, opened 2022-05-03
Add a Setting to force old style params download (no MavFtp)

EosBandi:
MavFtp needs a flawless comm link; otherwise, it can take longer to download the param file. This change adds a setting to force the use of old-fashioned param download.


meee1:
can you cleanup the mono change? ie exclude it

meee1:
also if you are having direct issues with mavftp, it would be good to chase it down, as the improvements are worth it.

there was a change  recently to do with packet size (because of radios), and also a burst read packet index issue

PR #2865, opened 2022-05-23
Using vswhere to get build env

buldo:
Original scripts supports only VS Community. But with vswhere scripts can use any modern VS in any edition


meee1:
need to remove the exe

buldo:
@meee1 why? As I uderstand from this repo https://github.com/microsoft/vswhere this exe has no dependencies and can be stored in repo.

There is alternative way with using this powershell module https://github.com/Microsoft/vssetup.powershell

I can rewrite current bat scripts to powershell if You allow this

meee1:
its included as part of the installer package on all VS installs
${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe

buldo:
@meee1 Fixed

build - Lib.ps1
```diff
@@ -1,5 +1,10 @@
-$msbuildPath = Split-Path (& "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe" -latest -requires Microsoft.Component.MSBuild -find MSBuild\Current\Bin\amd64\MSBuild.exe | Select-Object -First 1) -Parent
-$env:PATH = $msbuildPath + ';' + $env:PATH
+$installationPath = ./vswhere.exe -prerelease -latest -property installationPath
```
meee1, line 1: this need fixing

meee1: i also dont think this file needed any changes based on what it does

Issue #2876, opened 2022-05-30
Plan: add ATTITUDE_TIME to command list for Copters


rmackay9:
Copter-4.3 includes support for the MAV_CMD_NAV_ATTITUDE_TIME command in Auto (see PR https://github.com/ArduPilot/mavlink/pull/272 and  PR https://github.com/ArduPilot/ardupilot/pull/20851)

It would be nice to have this available in beta MP's list of available mission commands for Copter.  The arguments are listed below, we probably don't have room in the column headers for the units and other advice but I've written them anyway.

param1: time (sec)
param2: roll (deg)
param3: pitch (deg, +ve = lean back, -ve = lean forward)
param4: yaw (deg)
param5: climb rate (m/s)

rmackay9:
Fixed in beta so closing, thanks!

Issue #2864, opened 2022-05-23
Plan: SCRIPT_TIME command should be moved up in drop-down


rmackay9:
Copter/Rover-4.2 includes support the NAV_SCRIPT_TIME mission command. This item was kindly added to the drop-down with PR https://github.com/ArduPilot/MissionPlanner/pull/2825 but unfortunately the order is not ideal  It is currently down  with the DO_ commands instead of being up with the other NAV commands

![image](https://user-images.githubusercontent.com/1498098/169740135-98bb264d-0a3b-4f0b-9eb4-4374f7659ab0.png)

As a side note, I think that all the DO_ commands should be re-sorted into alphabetical order.  The CONDITION_ commands could all go at the bottom because they aren't used much.

rmackay9:
Fixed in beta so closing, thanks!

PR #2879, opened 2022-06-06
mavcmd: move SCRIPT_TIME up and add ATTITUDE_TIME

rmackay9:
This resolves two minor issues that I've raised recently

- https://github.com/ArduPilot/MissionPlanner/issues/2876
- https://github.com/ArduPilot/MissionPlanner/issues/2864

I have lightly tested this and the commands appear in the correct place.  The only issue is that Mavlink.cs doesn't contain ATTITUDE_TIME=42703 on my machine but I suspect a proper re-build using the latest ArduPilot/mavlink will resolve this.
image

### DIFF
--- a/mavcmd.xml
+++ b/mavcmd.xml
@@ -111,6 +111,15 @@
       <Y></Y>
       <Z></Z>
     </SCRIPT_TIME>
+    <ATTITUDE_TIME>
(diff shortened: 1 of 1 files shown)


rmackay9:
Thank you so much!

PR #2882, opened 2022-06-11
mavcmd: add DO_GIMBAL_MANAGER_PITCHYAW command

rmackay9:
This adds the new DO_GIMBAL_MANAGER_PITCHYAW command to the list of available mission commands.  Support for this command is being added to the flight code with PR https://github.com/ArduPilot/ardupilot/pull/20915 which will likely be merged within a week.

This has been lightly tested in SITL and seems to work OK.  There is still the issue of Param5 (aka "x") being corrupted but this is an existing issue that also occurs for ATTITUDE_TIME.

image

### DIFF
--- a/mavcmd.xml
+++ b/mavcmd.xml
@@ -12,60 +12,6 @@
       <Y>Long</Y>
       <Z>Alt</Z>
     </WAYPOINT>
-    <SPLINE_WAYPOINT>
-      <P1>Delay</P1>
-      <P2></P2>
-      <P3></P3>
-      <P4></P4>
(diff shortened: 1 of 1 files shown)


rmackay9:
I've also change the copter command list in these ways:

- removed the unimplemented CONDITION_CHANGE_ALT and DO_AUTOTUNE_ENABLE commands
- reordered the command list to be mostly alphabetical except for the top 3 most used navigation commands (waypoint, takeoff, return-to-launch)

I have lightly tested that the list looks OK
image

PR #2886, opened 2022-06-19
GCSViews: Change compile error CS1503

muramura:
I have corrected about CS1503.
I determined that it was a method mismatch because the argument was Image.
I am able to display on Android 7.0 and 12.0 with this fix.

error CS1503:
Screenshot from 2022-06-19 16-57-15


meee1:
please update your submodules, this is not a current problem

muramura:
@meee1 san.. Thanks for the comment, I updated SUBMODULE and built it.
The build finished successfully.

PR #2898, opened 2022-06-23
ParamCompare: Compare float values not strings so that matches are correct

hendjoshsr71:
This should fix the problem where the values are identical within float precision but show as different. 

Not sure if this would work in PX4 though.... since ArduPilot doesn't have strings in parameters.

Before PR
image

After PR:
image


meee1:
i will need to revisit this, as this was changed recently for other reasons

hendjoshsr71:
Okey doke. Any other details on that?  

Just annoys me so I'm trying to do my part to find solutions. 

Though I confess to knowing 0% C# so just try to muddle solutions to issues.

PR #2905, opened 2022-07-11
MavCmd: alphebetise and fixup rover command list

rmackay9:
This PR makes a few changes primarily to the Rover command list (there is one change to the Plane list)

- alphebetise the rover command list.  Like Copter's list the nav commands appear at the top with the two most common commands (WAYPOINT and RETURN_TO_LAUNCH) appearing right at the top.  Other less popular nav commands are then in alphabetical order.  "do" and "conditional" commands are next and are all in alphabetical order.
- removed unsupported do-set-home
- removed unsupported do-mount-configure from both rover and **plane**'s lists.
- added titles to the do-digicam-configure and do-digicam-control messages (for rover)
- removed unsupported arguments from do-change-speed (for rover)

I have tested this on my local machine by copying the mavcmd.xml and confirming the commands appear as expected

image

### DIFF
--- a/mavcmd.xml
+++ b/mavcmd.xml
@@ -837,15 +837,6 @@
       <Y></Y>
       <Z></Z>
     </DO_CHANGE_SPEED>
-    <DO_SET_HOME>
-      <P1>Current(1)/Spec(0)</P1>
-      <P2></P2>
-      <P3></P3>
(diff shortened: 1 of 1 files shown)


rmackay9:
Thanks!

PR #2907, opened 2022-07-19
Grab param defaults and display in full list

IamPete1:
Implementation of updated ftp format per https://github.com/ArduPilot/ardupilot/pull/20972

Displays all defualt values and adds tick box to only show params which have a none default value.
image

This is the full list only, still works with older version new stuff is hidden in that case.

Would be nice to make the save to file have a option to only save none defualts too.

Only draft because needs the AP pr merged first.


ExtLibs/ArduPilot/parampck.cs
```diff
@@ -118,10 +132,10 @@ private static object decode_value(char type_format, byte[] data, int startIndex
         {
             switch (type_format)
             {
-                case 'b': return data[startIndex];
-                case 'h': return BitConverter.ToUInt16(data, startIndex);
-                case 'i': return BitConverter.ToUInt32(data, startIndex);
-                case 'f': return BitConverter.ToSingle(data, startIndex);
+                case 'b': return (sbyte)data[startIndex];
+                case 'h': return (short)BitConverter.ToUInt16(data, startIndex);
```
meee1, line 136: whats the intent here?

ExtLibs/ArduPilot/parampck.cs
```diff
@@ -97,6 +100,16 @@ public static MAVLink.MAVLinkParamList unpack(byte[] data)
                 Array.Copy(data, dataPointer, vdata, 0, mapped.type_len);
                 dataPointer += mapped.type_len;
 
+                double default_value = Double.NaN;
```
meee1, line 103: can we change this to "decimal" from double.  the reason being a recent rounding issue, and decimal internally deals with this better.

ExtLibs/ArduPilot/parampck.cs
IamPete1: I was getting -1 warping to int max. This matches the return types from here:
https://github.com/ArduPilot/MissionPlanner/blob/43712f778807fcaa39527d3cfaaccd2ac4c9cbe5/ExtLibs/Mavlink/MAVLinkParam.cs#L48-L55

ExtLibs/ArduPilot/parampck.cs
IamPete1: changed to a nullable decimal `decimal?`

IamPete1:
AP support for this just go merged, I have also just retested with stable AP 4.2 and it still works with this change.

ExtLibs/ArduPilot/parampck.cs
meee1: sure. makes sence

ExtLibs/ArduPilot/parampck.cs
meee1: should we be casting? or reading in the correct format?  ie ToUint16 vs ToInt16 ?

IamPete1: Your right, it looks like I just have the wrong conversion. Copy and paste fail on my part.

meee1:
final confirmation. you happy with this now?

IamPete1:
@meee1 yep, I think so. Thanks.

PR #2914, opened 2022-07-20
GMapMarkerWP: Show WP Numbers Fixes #2913

### LINKED ISSUE #2913: display WP numbers on planner screen
when planning a mission it is frustrating that the WP numbers are not displayed on the planning screen. When you have a lot of waypoints (eg. 300 or more) then hunting for the WP is slow
can we always display the WP number in the green tear-drop on the planning screen?
image

hendjoshsr71:
Fixes it so that WP labels show up at all zoom levels.

I'm not sure how compatible this is with android or other screen size builds though.

### DIFF
--- a/ExtLibs/Maps/GMapMarkerWP.cs
+++ b/ExtLibs/Maps/GMapMarkerWP.cs
@@ -55,8 +55,7 @@ namespace MissionPlanner.Maps
             if (txtsize.Width > 15)
                 midw -= 4;
 
-            if (Overlay.Control.Zoom> 16 || IsMouseOver)
-                g.DrawImageUnscaled(fontBitmaps[wpno], midw,midh);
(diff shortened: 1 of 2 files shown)


meee1:
you do realize that you are undoing something that was done on purpose?   ie what happens when we get to over 99 wps?   or even 600 wps?

hendjoshsr71:
I think @tridge and I would want to know some of the constraints you have? It is quite hard for airplanes to plan landing/takeoffs when you cannot see the WP numbers.  There would also be issues for long distance missions where you are zoomed out quite far. 

By me understanding the WP symbols do not get larger/smaller when adding the numbers? So the issue isn't draw size?

Here are 200+ WPs thrown in.  The main issue to me is the symbol size doesn't go around the WP number. 
image


**Is it a drawing resource limitation due to android or slow computers?** 

Should we only display USER_SETTING WPs at time within the center of the screen?

PR #2921, opened 2022-08-05
Improve display of Battery and GPS data on HUD

sborenstein-ift:
Changes to the HUD add display improvements for the Battery and GPS displays in single and dual line modes, and move all text to be "bottom justified".


ExtLibs/Controls/HUD.cs
```diff
@@ -31,9 +31,9 @@
 
 namespace MissionPlanner.Controls
 {
-    public class HUD2: HUD
+    public class graphicsObject: HUD
```
meee1, line 34: can this be undone

ExtLibs/Controls/HUD.cs
```diff
@@ -102,6 +102,21 @@ private void FreeBitmap()
                 bitmap = null;
             }
         }
+
```
meee1, line 105: and this undone

meee1:
could you post a before and after screenshot?

sborenstein-ift:
image
image
image

Attached are normal 2-line, min-size 2-line and single line screen caps.

sborenstein-ift:
Here's the "before" image, single line and a quick battery 2-line. Note, in dual batt and dual GPS view, the lines have no space between. 
image
image

ExtLibs/Controls/HUD.cs
sborenstein-ift: Good question. I did not explicitly do this. I wonder if MVS took over here.

sborenstein-ift:
@meee1 I just reverted the first commit, which wasn't necessary, which addresses the questions you had. I think when I first started modifying this file, MVS did some wierd things to the RESX file. That wasn't necessary, I just removed that commit. Only the CS file is modified and changes should make more sense.

meee1:
> @meee1 I just reverted the first commit, which wasn't necessary, which addresses the questions you had. I think when I first started modifying this file, MVS did some wierd things to the RESX file. That wasn't necessary, I just removed that commit. Only the CS file is modified and changes should make more sense.

yeah, i was going to say the resx is redundant, as its 100% user drawn in code, so no designed parts.

PR #2926, opened 2022-08-18
Generates a plugin to test the Open Drone ID (ODID) 

sborenstein-ift:
Create a Open Drone ID Plugin that, when present, creates a Drone ID action tab, which contains the functionality. This includes two components: a NMEA GPS connection and a ODID UI that monitors the current ODID status. This implementation is for testing only, and is not yet ready to support full standard ID. 

Note: at this time, this plugin will likely only support a single connected MAV.


GCSViews/FlightData.Designer.cs
```diff
@@ -77,6 +77,8 @@ private void InitializeComponent()
             this.myButton1 = new MissionPlanner.Controls.MyButton();
             this.myButton2 = new MissionPlanner.Controls.MyButton();
             this.myButton3 = new MissionPlanner.Controls.MyButton();
+            this.tabDroneID = new System.Windows.Forms.TabPage();
```
meee1, line 80: given i know the plugin adds this, i dont think this should be here

ExtLibs/LEDBulb/Bulb/Bulb.csproj
```diff
@@ -0,0 +1,100 @@
...
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">x86</Platform>
+    <ProductVersion>8.0.30703</ProductVersion>
+    <SchemaVersion>2.0</SchemaVersion>
+    <ProjectGuid>{AF81C41B-06E6-4CA5-9FD4-A4E7042F1107}</ProjectGuid>
+    <OutputType>Library</OutputType>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>Bulb</RootNamespace>
+    <AssemblyName>Bulb</AssemblyName>
+    <TargetFrameworkVersion>v4.6.1</TargetFrameworkVersion>
```
meee1, line 13: can we fix this

C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Current\Bin\amd64\Microsoft.Common.CurrentVersion.targets(1220,5): error MSB3644: The reference assemblies for .NETFramework,Version=v4.6.1 were not found. To resolve this, install the Developer Pack (SDK/Targeting Pack) for this framework version or retarget your application. You can download .NET Framework Developer Packs at https://aka.ms/msbuild/developerpacks [D:\a\MissionPlanner\MissionPlanner\ExtLibs\LEDBulb\Bulb\Bulb.csproj]

meee1: ie 4.7.2

GCSViews/FlightData.Designer.cs
sborenstein-ift: Correct. I rebased the FlightData files back to master, but looks like I missed FD Designer.

sborenstein-ift: Should be removed from PR

ExtLibs/LEDBulb/Bulb/Bulb.csproj
sborenstein-ift: Just pushed fix.

sborenstein-ift:
FYI - Working on a few updates to this PR. Should have them pushed before AU wakes up.

sborenstein-ift:
This has been updated with the following fixes: 

- fixed timing for outgoing ODID msgs and added System Update
- fixed bug causing exception when connecting to bluetooth GPS
- added NMEA stream viewer for connected GPS 

Outstanding issues: 
- installing the DLL files alone is problematic on windows. This is just a windows issue. 
- I found a bug that causes added tabs in DLL plugins to not add a tab if there hasn't been a tabcontrolactions array set in the config.xml file. This is likely for new Mission Planner installs. The current workaround is to just open tab customize and tick one of the tabs on or off, then close Mission Planner and reopen. I will work on fixing this bug later on.

sborenstein-ift:
Added a fix for the Sats and HDOP display

Plugins/OpenDroneID2/NMEA_GPS_Connection.cs
```diff
@@ -0,0 +1,402 @@
...
+                init_com_port_list();
+            }
+            catch
+            {
+                Console.WriteLine("Couldn't Init Open DID Form.");
+            }
+            timer2.Start();
+        }
+
+        private void timer1_Tick(object sender, EventArgs e)
```
meee1, line 56: can we background thread this operation?
the reason is if for whatever reason we dont read a full line, or there is some serial io issue, the UI will hang for the timeout period.

sborenstein-ift: yes, that would be a fairly minor change. I can have it done before your tomorrow morning.

sborenstein-ift:
Updated to background operations and put place holder in for the tab fix. Will uncomment when FlightData is updated.

PR #2931, opened 2022-08-25
ODID Plugin - Set outgoing messages to be broadcast

sborenstein-ift:
This supercedes PR 2930, which Mostly fixes for the NMEA threading and auto-connect. Also adds a version number to the UI. this fix changes the ODID messages out to be broadcast component, to avoid routing issues on the AP end.


meee1:
just fyi, you can force push over an existing PR branch and github will deal with it

PR #2938, opened 2022-09-06
ODID - Create plugin based map overlay for the RID status

sborenstein-ift:
Creates an overlay status on connect to a ODID enabled device, onto the map. Includes a capability to double click either of the master indicators to declare an emergency.


sborenstein-ift:
This commit is version 0.04 of ODID plugin.

PR #2953, opened 2022-09-29
ExtLibs: Display units for horizontal and vertical accuracy

muramura:
Display units for horizontal and vertical accuracy.
Items without units were recognized as index values.
I learned that they are units of meters.
I shortened the item name because it was long.

Screenshot from 2022-09-30 06-19-31


rmackay9:
OK, interesting. I didn't even know that MP was using this file in the Quick display!

PR #2958, opened 2022-10-04
AuxOptions: add

meee1:


rmackay9:
Thank you very much!  I'm not personally able to build MP for some reason but I'm very happy to test a binary and/or the beta if this is included.  txs again!

meee1:
should be in beta shortly

meee1:
new beta again, fixing a bug that meant it didnt work

rmackay9:
I've tested this just now in beta and it works!! thanks so much, super useful!

rmackay9:
One small thing is we may want to remove the "Toggle" button which momentarily goes high and then low again.. maybe it's useful in some cases but the auxiliary functions are already edge triggered..

tridge:
@meee1 many thanks for this!
a couple of issues:
 - toggle does a rapid on/off. I don't think we want toggle at all for Aux functions
 - ideally it would remember the last Aux function sent, and when we see the DO_AUX_FUNCTION command ACK we display OK/Fail in HUD with the function name if the reply comes in within 2 seconds

PR #2959, opened 2022-10-04
Feature: Dropout forms for Flight Data and Flight Planner

LEGIONBOSS:
This feature makes it possible to drop out a various elements from the Flight Data screen into seperate windows.

Fetaures include:
 - Extendable dropout form management (new controls can be added as needed)
 - Under the HUD Tabs:
   - Dropout form functionality on a double click on the tab header (and off-session saving of it's position, size and dropped-out state)
    - Removal of "undock" action from Quick tab (has been replaced by the double click method)
    - Indexed in-order insertion of tabs (when their dropout forms are closed)
 - Flight Planner: Dropout form functionality with the same off-session saving and loading
   - (the dropout window closes if the user swithes to the Flight Planner window using menu strip at the top)
 - HUD: Off-session saving and loading of positoon, size and dropped-out state
 - Connection Stats Window: Off-session saving and loading of positoon, size and dropped-out state

All saving and loading is handled using the already existing functions from the `ControlHelpers` and `Settings` classes.


EosBandi:
This is a very helpful change for multi-monitor GCS systems.  We started to use it a while ago.. (Notice, LegionBoss is working in my team :) )

PR #2967, opened 2022-10-08
Reboot: auto reconnect

IamPete1:
This auto reconnects after a sending a reboot command, useful when doing initial setup where lots of reboots are needed. Also handy for catching very early start up messages. 

Testing:
- direct USB reconnects
- USB telemetry does not reconnect because the link is not closed
- TCP to SITL does not reconnect (it would be nice to get this working in the future, maybe SITL does not close the link correctly)
- firmware upload works as normal

### DIFF
--- a/ExtLibs/ArduPilot/Mavlink/MAVLinkInterface.cs
+++ b/ExtLibs/ArduPilot/Mavlink/MAVLinkInterface.cs
@@ -2407,12 +2407,27 @@ Mission Planner waits for 2 valid heartbeat packets before connecting");
             {
                 var ans1 = doCommand((byte) sysidcurrent, (byte) compidcurrent, MAV_CMD.PREFLIGHT_REBOOT_SHUTDOWN,
                     param1, 0, 0, 0, 0, 0, 0);
-                if (ans1)
-                    return true;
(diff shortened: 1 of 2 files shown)


ExtLibs/ArduPilot/Mavlink/MAVLinkInterface.cs
```diff
@@ -2407,12 +2407,27 @@ public bool doReboot(bool bootloadermode = false, bool currentvehicle = true)
...
-                if (ans2)
+                }
+                // Successful reboot
+                if (ans1 || ans2)
+                {
+                    // Not going into bootloader and on a serial connection
+                    if (!bootloadermode && (BaseStream is SerialPort))
+                    {
+                        // Direct USB will disconnect after a reboot, wait and see if we should re-connect
+                        Thread.Sleep(500);
```
IamPete1, line 2423: This timeout seems to work, its bit of a guess, not sure if it would work in every case, maybe there is a better way?

ExtLibs/ArduPilot/Mavlink/MAVLinkInterface.cs
```diff
@@ -2407,12 +2407,27 @@ public bool doReboot(bool bootloadermode = false, bool currentvehicle = true)
...
+                        if (!BaseStream.IsOpen)
+                        {
+                            Open(true);
+                        }
```
IamPete1, line 2427: We could auto refresh params if the link still open. That would be the right thing in most cases but quite annoying on a very slow link.

PR #2971, opened 2022-10-09
FlightMode: Set flight mode setting to setup only

muramura:
The flight mode settings are located on the Setup and Configure screen.
I would prefer one screen.
I removed it from the config screen.

AFTER
Screenshot from 2022-10-10 00-48-53

BEFORE
flight-002

flight-001

### DIFF
--- a/GCSViews/SoftwareConfig.cs
+++ b/GCSViews/SoftwareConfig.cs
@@ -62,11 +62,6 @@ namespace MissionPlanner.GCSViews
                 {
                     if (MainV2.comPort.BaseStream.IsOpen)
                     {
-                        if (MainV2.DisplayConfiguration.displayFlightModes)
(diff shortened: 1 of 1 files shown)


GCSViews/SoftwareConfig.cs
```diff
@@ -62,11 +62,6 @@ private void SoftwareConfig_Load(object sender, EventArgs e)
                 {
                     if (MainV2.comPort.BaseStream.IsOpen)
                     {
-                        if (MainV2.DisplayConfiguration.displayFlightModes)
-                        {
-                            start = AddBackstageViewPage(typeof(ConfigFlightModes), Strings.FlightModes);
```
meee1, line 67: we  might need to set a new "start"  just to set the default on first load

PR #2974, opened 2022-10-11
ConfigMount: Change to group and item names

muramura:
The other parameter names are group and item names.
I want the group name and the item name.
This change will be an obstacle when supporting MNT2.

AFTER
Screenshot from 2022-10-12 07-13-38

Screenshot from 2022-10-12 07-14-14

### DIFF
--- a/GCSViews/ConfigurationView/ConfigMount.cs
+++ b/GCSViews/ConfigurationView/ConfigMount.cs
@@ -97,8 +97,8 @@ namespace MissionPlanner.GCSViews.ConfigurationView
                 }
             }
 
-            CMB_mnt_type.setup(ParameterMetaDataRepository.GetParameterOptionsInt("MNT_TYPE",
-                MainV2.comPort.MAV.cs.firmware.ToString()), "MNT_TYPE", MainV2.comPort.MAV.param);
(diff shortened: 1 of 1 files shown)


rmackay9:
This looks like an improvement although there are lots of other changes we need to make to this page to properly support the AP flight code enhancements in AP-4.3.

- MNT_ renamed to MNT1_
- Angle limits renamed to MNTx_ROLL_MIN/MAX, MNTx_PITCH_MIN/MAX, MNTx_YAW_MIN/MAX and changed from centi-degrees to degrees
- MNT_STAB_xx parameters removed (users can now choose whether the gimbal is stabilized or not using the MNTx_TYPE)
- MNT_JSTICK_xx renamed to MNTx_RC_RATE
- MNT_ROLL_RC_IN (or similar) has been replaced with RCx_OPTION (Mount Roll).  Same for Pitch and yaw
- Control angles should be removed (these disappeared a few versions ago actually)
- In general we should use "Pitch" and "Yaw" instead of "Tilt" and "Pan".  We don't need to use different terminology for the gimbal compared to the main flight controller.

muramura:
@rmackay9 san.
This PR change is also necessary when changing MNT to MNT1 and MNT2.

PR #2979, opened 2022-10-14
Set the low level of SEVERITY to speak

muramura:
I set the voice notification and HUD display level to NOTICE or higher.
There are some messages I would like to be notified about in the voice notification and HUD display.
I know that I am voice notified of NOTICE or above in QGC.
ArduPilot is notifying me of NOTICE level messages.
I think there is merit to have this message displayed on the HUD, with voice notification.

Other:.
The message display on the HUD uses red, yellow, and white colors to distinguish between emergency, warning, and other.

### DIFF
--- a/ExtLibs/ArduPilot/Mavlink/MAVLinkInterface.cs
+++ b/ExtLibs/ArduPilot/Mavlink/MAVLinkInterface.cs
@@ -5221,7 +5221,7 @@ Mission Planner waits for 2 valid heartbeat packets before connecting");
                         if (MAVlist[sysid, compidcurrent].cs.version.Major > 0 ||
                             MAVlist[sysid, compidcurrent].cs.version.Minor >= 4)
                         {
-                            if (sev <= (byte)MAV_SEVERITY.WARNING)
+                            if (sev <= (byte)Settings.Instance.GetInt32("severity", 4))
                             {
                                 printit = true;
                                 mavsev = (MAV_SEVERITY)sev;
--- a/ExtLibs/Utilities/DisplayView.cs
+++ b/ExtLibs/Utilities/DisplayView.cs
@@ -17,6 +17,19 @@ namespace MissionPlanner.Utilities
         Advanced,
         Custom
     }
+
+    public enum SeverityLevel
+    {
+        Emergency,
+        Alert,
+        Critical,
+        Error,
+        Warning,
+        Notice,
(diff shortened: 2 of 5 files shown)


muramura:
I did an AUTOTUNE, saw that the aircraft went quiet, and looked at the MESSAGE tag to see what caused it.
I could see that ArduPilot was notifying me with NOTICE.
NOTICE is the audio notification level.
I would suggest that voice notification to the operator would be a good idea.

Screenshot from 2022-10-27 06-21-14

meee1:
can you make this a user selectable option?

muramura:
I would add to the PLANNER configuration item.

muramura:
I have made it possible to set a SEVERITY minimum level.

ExtLibs/ArduPilot/Mavlink/MAVLinkInterface.cs
```diff
@@ -5221,7 +5221,7 @@ public async Task<MAVLinkMessage> readPacketAsync()
                         if (MAVlist[sysid, compidcurrent].cs.version.Major > 0 ||
                             MAVlist[sysid, compidcurrent].cs.version.Minor >= 4)
                         {
-                            if (sev <= (byte)MAV_SEVERITY.WARNING)
+                            if (sev <= (byte)Settings.Instance.GetInt32("severity"))
```
meee1, line 5224: can you add the default value. as if this has never been set by the user, it will results in a 0 atm

muramura: I would add processing-ization.

meee1, line 5224: what i was suggesting was
Settings.Instance.GetInt32("severity", 4)

muramura:
@meee1 san.
I have added processing-ization.
 https://github.com/ArduPilot/MissionPlanner/pull/2979/files#diff-aa20c6da1c0275948fc5a8466b97b615d01eb502ef06bbbaea4d31d407d6f1ddR92

ExtLibs/ArduPilot/Mavlink/MAVLinkInterface.cs
muramura: I figured.  I will change my suggestion.

PR #2989, opened 2022-10-27
use MAV_CMD_REQUEST_MESSAGE to get AUTOPILOT_VERSION

peterbarker:


meee1:
i understand the change, but what's going on to need now 3 methods to request a single message in response?

peterbarker:
On Wed, 26 Oct 2022, Michael Oborne wrote:

> i understand the change, but what's going on to need now 3 methods to
> request a single message in response?

Well.  Originally we got a message AUTOPILOT_VERSION (2014)

Then people wanted to prompt the autopilot for it 
(AUTOPILOT_VERSION_REQUEST) (2015)

Then people discovered that sucked and we have commands for a reason, so 
we got MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES (*also* 2015, but a day 
later than above...).

Then we discovered (early 2019) that we could really just set things up so 
you can request an instance of a message using MAV_CMD_REQUEST_MESSAGE), 
making both of those redundant.  Hamish deprecated a whole bunch of 
messages that simply prompted the autopilot for an instance of a message 
later that year.

If you have a look at the ArduPilot request you can see all the older 
things did was queue an AUTOPILOT_VERSION for sending....

BTW, I downloaded the artifact from github and the message request does 
arrive.

PR #2990, opened 2022-11-02
added plane-3d sitl target

tridge:
for aerobatics


GCSViews/SITL.resx
```diff
@@ -2436,6 +2436,9 @@
   <data name="cmb_model.Items32" xml:space="preserve">
     <value>rover-skid</value>
   </data>
+  <data name="cmb_model.Items32" xml:space="preserve">
```
meee1, line 2439: dup number

tridge: fixed, sorry

PR #2992, opened 2022-11-06
FEATURE: support NOT reading params from a device on-connect.

davidbuzz:
Adds a user checkbox in the 'Planner' setup screen[ defaults to ticked, so the default dehaviour is unchanged], and prevents params from being read fro mthe drone on-connect when this tickbox is unselected.


ExtLibs/ArduPilot/Mavlink/MAVLinkInterface.cs
```diff
@@ -307,7 +307,7 @@ public MAVState MAV
             set { MAVlist[sysidcurrent, compidcurrent] = value; }
         }
 
-        public double CONNECT_TIMEOUT_SECONDS = Settings.Instance.GetDouble("CONNECT_TIMEOUT_SECONDS", 30.0);
+        public double CONNECT_TIMEOUT_SECONDS = Settings.Instance.GetDouble("CONNECT_TIMEOUT_SECONDS", 9999.0);
```
magicrub, line 310: why the long timeout?

meee1: i agree, default should stay the same, you can override via the setting

davidbuzz: the long timeout is due to the client in  question needing to be able to "connect" to their rfd900 radio while the vehicle at the other end is out-of-range, and have the vehicle auto-connect immediately when it comes in-range, it should probably be both configurable and a separate PR, but i haven't got the yet.   marking this PR as WIP/draft.

PR #3006, opened 2022-11-19
Adds stabilize delay for digicam_control Trigger Method to survey(Grid)

JoeKae:
# Auto WP: Survey (Grid)

## Feature

When the trigger method is set to `DO_DIGICAM_CONTROL`, the flight plan has a waypoint for each cam shot. This PR adds the feature to include `DELAY` commands for each cam shot waypoint.

## Motivation

If the aircraft stops for each cam shot, a higher airspeed can be used. Also, stopping can improve image quality.

## For what type of aircraft?

- Multirotor
- Heli
- VTOL (maybe?)

stabilize_gui
stabilize_wp_list


JoeKae:
Was already implemented in grid options

PR #3007, opened 2022-11-22
GCSViews: Configuration: RadioOutput: allow upto 32 servos if  enabled

IamPete1:
Allows configuration of the full 32 servos from the Servo Output page if enabled.

image


meee1:
i totally get it, but that's crazy seeing it

PR #3013, opened 2022-12-01
Setup/Camera Gimbal update for 4.3

rmackay9:
This updates the Camera Gimbal setup screen so that it works for AP 4.3 but also sadly makes it incompatible with 4.2 (and earlier).  I think it is more important that MP is compatible with our stable build but if you have advice on how to make it backwards compatible and I'll try to do that.

Changes included are:

1. remove the CONTROL_X/Y/Z section (these parameters haven't existed for years)
2. remove the STAB_TILT/ROLL/PAN parameters.  These were required previously to allow users to use the "Servo" gimbal type for both regular servo gimbals and brushless gimbals (that self stabilize).  In AP 4.3 the "Servo" gimbal type is really only for servo gimbals.  For brushless-self-stabilising gimbals users should use the "BrushlessPWM" type
3. rename parameter prefix from "MNT" to "MNT1"
4. rename angle limit parameters from "ANGMIN_TIL" to "PITCH_MIN", "ANGMIN_ROL" to "ROLL_MIN", "ANGMIN_PAN" to "YAW_MIN", and same for the corresponding MAX parameters
5. cosmetic changes including
    - renamed "Tilt" label to "Pitch", "Pan" to "Yaw", "Trigger" to "Camera Trigger"
    - renamed "Servo Limits" labels to just "Servo"
    - moved servo drop downs to just above servo limit fields
    - slightly shrunk all images
    - moved all section titles to just below section's top line

I'm happy for the last commit (the cosmetic changes) to be dropped if you'd prefer.  I think the screen looks better but I don't feel strongly about it.
mp-gimbal-setup-PR-before-vs-after

BTW, I could not figure out how to make the "Input Ch" drop-downs work.  Any advice on how to fix these is also greatly appreciated.

I tested this on my own machine and it seemed fine.  I also lightly tested it with Copter-4.2 and as expected it does not allow the gimbal to be setup but it also doesn't throw out any ugly errors.
mp-gimbal-setup-PR-42-test

### DIFF
--- a/GCSViews/ConfigurationView/ConfigMount.cs
+++ b/GCSViews/ConfigurationView/ConfigMount.cs
@@ -149,10 +149,6 @@ namespace MissionPlanner.GCSViews.ConfigurationView
                 updateRoll();
                 updateYaw();
 
-                CHK_stab_tilt.setup(1, 0, ParamHead + "STAB_TILT", MainV2.comPort.MAV.param);
-                CHK_stab_roll.setup(1, 0, ParamHead + "STAB_ROLL", MainV2.comPort.MAV.param);
-                CHK_stab_pan.setup(1, 0, ParamHead + "STAB_PAN", MainV2.comPort.MAV.param);
-
                 NUD_NEUTRAL_x.setup(-180, 180, 1, 1, ParamHead + "NEUTRAL_X", MainV2.comPort.MAV.param);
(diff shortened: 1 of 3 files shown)


rmackay9:
I think I will close this for now because it is unable to be merged due to it breaking compatibility with older versions of AP.

PR #3017, opened 2022-12-07
ConfigMount: remove Control and Pan stabilize controls

rmackay9:
This is a small enhancement to the mount setup page to fix up two very obvious problems.  This is a subset of this larger PR https://github.com/ArduPilot/MissionPlanner/pull/3013.  This doesn't allow this page to be used for the stable releases (4.3).. this will come in a follow-up PR.  I created this small PR in the hopes that it would be easier to review and merge.

1. remove the CONTROL_X/Y/Z section (these parameters haven't existed for years)
2. remove the STAB_PAN parameter checkbox.  Users should never be setting this parameter so it is best that we just remove it.

This has been tested on my PC connecting to a Pixhawk1 running Copter-4.2.3.


rmackay9:
Thanks!

PR #3035, opened 2023-01-09
ExtLibs: Allow for special characters in NTRIP connection string

stephendade:
If a user's ntrip username or password has special characters (not letters or numbers), ie ``http://user1#:938aasd^!@ntrip.data.gnss.ga.gov.au:2101/TEST1`` the URI string will not be properly encoded.

This PR allows for these special characters to be used, which is useful as some NTRIP providers are now enforcing password policies of requiring special characters.

### DIFF
--- a/ExtLibs/Comms/CommsNTRIP.cs
+++ b/ExtLibs/Comms/CommsNTRIP.cs
@@ -113,8 +113,29 @@ namespace MissionPlanner.Comms
             Open(url);
         }
 
+        public static string PercentEncode(string value)
+        {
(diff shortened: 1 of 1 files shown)


meee1:
i havnt confirm the function
but wouldusing
WebUtility.UrlEncode(item.Value)
be an option?

stephendade:
>  WebUtility.UrlEncode(item.Value)

That encodes the whole string: ``http%3A%2F%2Fuser1%23%3A938aasd%5E!%40ntrip.data.gnss.ga.gov.au%3A2101%2FTEST1``

PR #3043, opened 2023-01-14
GCSViews: ConfigurationView: Motor test: use JSON file to add more infromation

IamPete1:
Firstly this grabs the frame type and class from param meta data. Secondly it disaplays the motor number and rotation direction by looking it up in a json file (which I have not worked out how to include correctly).

image

With more motors:

image

Ultimately I would like to programmatically generate the diagram, but that is a bit too much for me. I would like to end up with a interface that will re-odder your motors correctly, you would test motor A and then click on the diagram which motor actually spun and it would switch them round until your ordering is correct. 

The json file is generated by a modified AP motors example, it only has all the matrix types. We could update it to include tricopters and other unusual frame types by hand.


GCSViews/ConfigurationView/ConfigMotorTest.cs
```diff
@@ -178,12 +198,71 @@ private int get_motormax()
...
+
+            return true;
+        }
+
+
+        private void lookup_frame_layout(int frame_class, int frame_type)
+        {
+            try
+            {
+                using (StreamReader r = new StreamReader("C:/motor_layout.json"))
```
IamPete1, line 239: This is not the correct way to do this. I could not work out how to include the file with the build.

meee1: you can use this a a reference
https://github.com/ArduPilot/MissionPlanner/blob/6f95f65707e15b8e412432cc33aee22aa4daa868/Utilities/Firmware.cs#L124

meee1: https://github.com/ArduPilot/MissionPlanner/blob/5dac1d51464f282735bf8d5d4bf7769ffa09c429/MissionPlanner.csproj#L4935

meee1: need to add it to missionplannerlib as well

IamPete1: done, thanks.

IamPete1:
Rebased and now includes the json file correctly, I might have gone overboard and included it in more places than needed, but it seems to work.

PR #3046, opened 2023-01-21
Log review: support bitmask fields from metadata

IamPete1:
This adds support for parsing bitmask masks from the logmetadata xml and applying them to values. So you can plot each bit field individualy. Currenlly the TECS message is the only example in the xml, because of this I have not tested instance with bitmask, but it should work.

image

image

As a byproduct we also now get mask with `&` as a manual modifier too. The modifiers also are included in the label, for example:

image

We need to update the log meta data a little to give a name and description for each bit. I would also like to give a mask rather than a bit so we can do multiple bit values (there are some in the EKF I think). Maybe we should do that metadata rework before merging this, @peterbarker any thoughts?

### DIFF
--- a/ExtLibs/ArduPilot/LogMetaData.cs
+++ b/ExtLibs/ArduPilot/LogMetaData.cs
@@ -25,7 +25,19 @@ namespace MissionPlanner.ArduPilot
 
         static string url = "https://autotest.ardupilot.org/LogMessages/{0}/LogMessages.xml.xz";
 
-        public static  Dictionary<string, Dictionary<string, string>> MetaData { get; } = new Dictionary<string, Dictionary<string, string>>();
+        public struct LogItemFeild
+        {
+            public string description;
+            public struct LogItemFeildBitmask
+            {
+                public string name;
+                public uint mask;
+                public string description;
+            }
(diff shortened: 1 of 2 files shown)


peterbarker:
I would *really* prefer to get https://github.com/ArduPilot/ardupilot/pull/13858 in.  That PR was initially bounced because people didn't want to try to parse the C++ code to extract the enumerations.  Despite the fact that it only failed on 1 or 2 in our entire codebase.... (see https://github.com/ArduPilot/ardupilot/pull/13858/files#diff-85edce785b34ddd0e30321128f2a304e9660369e408f7df4e36542cc8a0d2feeR81)

This is the critical line; we reference the extracted enumeration value in the parameter metadata by name: https://github.com/ArduPilot/ardupilot/pull/13858/files#diff-8423e07ce80b3c668160f28a03b5969a4a6b137bd405eb19fa80fc03156fef48R80

Should I dust that PR off?

IamPete1:
Thanks to https://github.com/ArduPilot/ardupilot/pull/13858 and https://github.com/ArduPilot/ardupilot/pull/22823 we now have a better way to add bitmasks to the xml.

image

Here on a instance, raw value 6.

image

Made up of active 

image

and streaming

image

PR #3048, opened 2023-01-27
ConfigFirmware: allow offline access to certain parts

robertlong13:
Fixes an issue that Alex/ARACE pointed out on Discord. There are circumstances where someone might want to upload custom firmware with a computer that has no internet access at the time.

image


rmackay9:
That was quick!

PR #3057, opened 2023-02-10
HUD prearm status

robertlong13:
Add an indicator to the HUD to display the prearm status of the aircraft.
image

Clicking on the indicator brings up a window that displays the failing prearm checks. While the window is open, the GCS sends `RUN_PREARM_CHECKS` messages every 5 seconds to ensure fresh failure messages in the window. When the aircraft is armed, the indicator is hidden, and any open detail window gets closed.

The 5-second request frequency means I needed to silence the speech specifically for prearm messages (arm failure messages still speak out loud). I think this is a fair thing to do, as the new indication is far easier to work with than the audible warning, and I have always thought the prearm messages should be silently displayed, as even at the 30s interval, they are annoying. I can make this silencing a configurable option though if we think this will be controversial.

Funded by ARACE


pompecukor:
My team and I teasted. It turned out even better than we originally hopped. Well done @robertlong13 .
Please merge as soon as possible. I am very confident that the entire community will find this useful.

tridge:
very nice! Thanks

peterbarker:
Nice!

Controls/PrearmStatus.cs
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        private void requestchecktimer_Tick(object sender, EventArgs e)
+        {
+            if (MainV2.comPort.MAV.cs.prearmstatus) return; // Don't request prearm checks if we are ready to arm
+
+            lastRequestTime = DateTime.Now;
+
+            // Request prearm checks to be performed
+            MainV2.comPort.doCommand(
```
meee1, line 35: please handle exceptions on this

Issue #3062, opened 2023-02-23
Support Link button


tridge:
I've been using https://support.ardupilot.org a lot to help users. I've found that many users struggle to follow the instructions.
Common mistakes are:
 - don't click write access
 - don't choose UDP client
 - worry about baudrate, as MP forces 4800 and users worry that is wrong
 - mis-spell support.ardupilot.org

A lot of the time I need to get the user to screen share to find their mistake. It is very frustrating as it is often a new user, and sometimes there are language issues and bandwidth is sometimes too low for screen share.

I would really appreciate a button under Setup/Advanced labelled "Support Link" or similar. It should bring up a dialog with 2 lines:

 - the server, which should be pre-filled with a default of support.ardupilot.org
 - the port, which should be labelled as 5 digit support ID

I'm hoping this will make it easier for end users, and save a lot of frustration with trying to work out why they can't get support

tridge:
@robertlong13 thanks for implementing!

PR #3065, opened 2023-03-01
ConfigAdvanced: add support.ardupilot.org proxy

robertlong13:
image

See https://github.com/ArduPilot/MissionPlanner/issues/3062

@tridge, any comments on the names and descriptions in that screenshot?


robertlong13:
The failed check is unrelated to this PR, just coincidental timing. [I get the same failure error when building master](https://github.com/robertlong13/MissionPlanner/actions/runs/4327697531). I will fix that at some point with another PR

PR #3066, opened 2023-03-02
Switched to AsyncKeyedLock

MarkCiliaVincenti:
This helps clean up from the dictionary and reduces memory allocations. However I'm not sure if `urlcacheSem` should be static. It feels like it should, but the existing code keeps the dictionaries separate.

### DIFF
--- a/Utilities/Firmware.cs
+++ b/Utilities/Firmware.cs
@@ -1,4 +1,5 @@
-﻿using log4net;
+using AsyncKeyedLock;
+using log4net;
 using ManagedNativeWifi.Simple;
 using MissionPlanner.Arduino;
 using MissionPlanner.ArduPilot;
@@ -253,8 +254,11 @@ namespace MissionPlanner.Utilities
             }
         }
 
-        object urlcachelock = new object();
-        Dictionary<string, SemaphoreSlim> urlcacheSem = new Dictionary<string, SemaphoreSlim>();
+        private AsyncKeyedLocker<string> urlcacheSem = new AsyncKeyedLocker<string>(o =>
+        {
+            o.PoolSize = 20;
+            o.PoolInitialFill = 1;
+        });
         static Dictionary<string, string> urlcache = new Dictionary<string, string>();
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


Utilities/Firmware.cs
```diff
@@ -305,15 +309,9 @@ void getAPMVersion(object tempin)
...
-                    urlcacheSem[url.AbsoluteUri] = new SemaphoreSlim(1, 1);
-
-            urlcacheSem[url.AbsoluteUri].Wait();
-
-            try
+            using (urlcacheSem.Lock(url.AbsoluteUri))
             {
-                lock (urlcachelock)
+                lock (urlcacheSem)
                     if (urlcache.ContainsKey(url.AbsoluteUri))
```
meee1, line 315: is this guaranteed to contain the key now? given the above using.

i understand that your the author of the library?

the intent of this function was to dedup http requests/cache and respond

MarkCiliaVincenti: I'm not sure I follow what you mean in your first sentence.

Yes, I am the author of the AsyncKeyedLock library.

meee1: line 
using (urlcacheSem.Lock(url.AbsoluteUri))

must create a state
that
 if (urlcache.ContainsKey(url.AbsoluteUri))
 
 will not fail on? ie is it redundant?

MarkCiliaVincenti: No I don't think so. Imagine you fire up the application, urlcache and urlcacheSem will both be empty. If you make a new call to a URL, urlcacheSem will be used, obtains the lock and goes check urlcache to see if it's in the cache. It won't find it.

MarkCiliaVincenti:
Is there anything else you need @meee1 ?

MarkCiliaVincenti:
@meee1  are you interested or shall I close this PR?

MarkCiliaVincenti:
@meee1

MarkCiliaVincenti:
@meee1  please let me know

MarkCiliaVincenti:
@meee1

PR #3068, opened 2023-03-06
Upgrade projects to net472

hexate:
Upgrade .Net to net472 as net461 is no longer supported.


hexate:
I removed or updated all references to net461 as it is no longer supported by Microsoft.

There were several projects that were already updated to net472 but were still writing build files to net461 directories, so I renamed those targets also.

PR #3078, opened 2023-03-14
MyButton: remove "Browsable" from TextColorNotEnabled

robertlong13:
This is not necessary. It clutters up designer files and can cause backward-compatibility issues for plugins designed after that was added.

Not a big deal at all, but it was bothering me a little.


robertlong13:
Never mind, this doesn't actually stop the designer from throwing the garbage in there. Oh well.

PR #3080, opened 2023-03-15
Param list improvements

robertlong13:
The really tall rows caused by long value descriptions makes the full param list hard to look at. Additionally, the cells causing this weren't particularly easy to read. Instead of auto-fitting the row height, I have opted to improve the look of the tooltip for overflowing cells. It is now significantly more readable.

image

While I was trying to bring some of those improvements over to the Full Parameter Tree view, I decided to overhaul it to make the two pages consistent. I think it looks and works much better to have a tree view separated in a split pane from the grid view. I'm pretty happy with how it turned out, and it could easily be combined into a single page instead of being offered as two separate options.


meee1:
the only thing to be mindfull of is Tree is not cross platform friendly. list is.

meee1:
although just knoticed you removed
BrightIdeasSoftware

so it might be friendly now

robertlong13:
I've actually decided to go all the way and eliminate the dedicated tree tab in favor of the merged set. I also added a button to collapse/expand the tree.
image

meee1:
> @meee1, any idea why I'm getting this issue on the Android build?
> 
> ```
> [Build Release APK](https://github.com/ArduPilot/MissionPlanner/actions/runs/4429522904/jobs/7770085962#step:1:41)
> You are running out of disk space. The runner will stop working when the machine runs out of disk space. Free space left: 11 MB
> ```
> 
> I can't imagine that's related to this PR.

fixed, cleared the build cache

robertlong13:
Added comboboxes, numeric up/down, and a button for launching the bitmask editor, to the options column.

image

PR #3090, opened 2023-04-12
Air-Link: Added support for connections to air-link modules

main2stels:
Hi. I am one of the developers of lte [modules](https://ardupilot.org/plane/docs/common-airlink-telemetry.html) for telemetry transmission . This add-on can enhance the functionality and usefulness of Mission Planner software for users to rely on LTE connectivity in their missions.


meee1:
> Please get the mavlink changes into ardupilot Mavlink and mavlink. I always follow master in this regard

PR #3099, opened 2023-04-22
CoT/TAK: add detail takv and some cleanup

magicrub:
CoT/TAK: add detail takv and some cleanup

### DIFF
--- a/Controls/SerialOutputCoT.cs
+++ b/Controls/SerialOutputCoT.cs
@@ -269,6 +269,11 @@ namespace MissionPlanner.Controls
             return "NOsysid" + sysid;
         }
 
+        bool isValidStr(object obj)
+        {
+            return (obj != null && obj.ToString().Length > 0);
+        }
+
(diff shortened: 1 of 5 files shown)


Controls/SerialOutputCoT.cs
```diff
@@ -269,6 +283,11 @@ string FindUIDviaSysid(byte sysid)
             return "NOsysid" + sysid;
         }
 
+        bool isValidStr(object obj)
+        {
+            return (obj != null && obj.ToString().Length > 0);
```
meee1, line 288: https://learn.microsoft.com/en-us/dotnet/api/system.string.isnullorempty?view=net-8.0

magicrub: Thanks for the tip. I was just refactoring your code though :)

`if (callsign != null && callsign.ToString().Length > 0 && endpoint != null && endpoint.ToString().Length > 0`

PR #3101, opened 2023-04-25
Jump Tags - add ability to Trigger it

magicrub:
This allows the GCS from the FlightData screen to jump to a JUMP_TAG. This PR is on top of @rmackay9 's PR  https://github.com/ArduPilot/MissionPlanner/pull/3097 which adds it to the mission pulldown list

### DIFF
--- a/GCSViews/FlightData.cs
+++ b/GCSViews/FlightData.cs
@@ -6181,5 +6181,29 @@ namespace MissionPlanner.GCSViews
             tabControlactions.Multiline = !tabControlactions.Multiline;
             Settings.Instance["tabControlactions_Multiline"] = tabControlactions.Multiline.ToString();
         }
+
+        private void jumpToTagToolStripMenuItem_Click(object sender, EventArgs e)
+        {
+            string tag_str = "";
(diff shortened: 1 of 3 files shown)


GCSViews/FlightData.cs
```diff
@@ -6181,5 +6181,29 @@ private void multiLineToolStripMenuItem_Click(object sender, EventArgs e)
...
+            if (!UInt16.TryParse(tag_str, out tag) || tag < 0 || tag > 0xFFFF)
+            {
+                CustomMessageBox.Show("Invalid Tag. Must be a number from 0 to 65535");
+                // NOTE: This is recursive to automatically re-pop up the dialog box
+                // on input error for as many times as you try to enter an invalid number.
+                jumpToTagToolStripMenuItem_Click(null, null);
+                return;
+            }
+
+            if (!MainV2.comPort.doCommand(MAVLink.MAV_CMD.DO_JUMP_TAG, tag, 0, 0, 0, 0, 0, 0))
```
meee1, line 6203: might need a try catch around this, its likerly to cause an exception

magicrub: fixed up with an additional commit

PR #3111, opened 2023-04-29
Param Out-Of-Range warning: after 3 YESes ask to assume yes

magicrub:
Param Out-Of-Range warning: after 3 YESes ask to assume yes


magicrub:
I haven't tested this much... yet

PR #3129, opened 2023-06-03
GCSViews: ConfigRawParams: bitmask takes priority over options

IamPete1:
Displaying the options drop down for bitmask parameters makes it impossible to set more than one option. We should probably not have have both in the param meta data, but there are some with both currently. `PILOT_THR_BHV` is a example of this on copter. Currently it will display the drop down:

image

With this PR we now get the bitmask button.

image


IamPete1:
@robertlong13 tagging you because I don't have permissions to request your review via GH.

robertlong13 (approved):
Looks good to me, and bitmask > combo makes sense as a priority (never anticipated both being in the meta when I wrote this)

PR #3130, opened 2023-06-03
GCSViews: ConfigurationView: update coloum width

IamPete1:
This messes with the param editor column widths to try and make resizing better. Param names and units change to a fixed width to fit everything ("AllCells").  This should mean that the other column's get more room when resizing. The "FAV" column is made none re-sizable, so it should stay small.

The remaining columns that can be changed are the value, default, options and description.


rmackay9:
Maybe this PR will resolve this issue I raised https://github.com/ArduPilot/MissionPlanner/issues/3134

robertlong13 (approved):
Looks good to me. Tested it out as well. I really like it.

PR #3139, opened 2023-06-26
mavcmd: update do-digicam-control fields

rmackay9:
This renames the "On/Off" field of the DO_DIGICAM_CONTROL command to "Session Control" which better matches [the mavlink spec](https://github.com/ArduPilot/mavlink/blob/master/message_definitions/v1.0/common.xml#L1226) and, I hope, will reduce confusion with the "Shutter Control" field which is what most users actually need to set.

This has been lightly tested on my desktop and the before and after shots are shown below.

BEFORE:
image

AFTER:
image

Thanks to @CraigElder for reporting this issue.

### DIFF
--- a/mavcmd.xml
+++ b/mavcmd.xml
@@ -193,7 +193,7 @@
       <Z></Z>
     </DO_DIGICAM_CONFIGURE>
     <DO_DIGICAM_CONTROL>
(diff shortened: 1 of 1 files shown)


rmackay9:
Thanks!

PR #3144, opened 2023-07-05
CommsUDPSerialConnect multicast to allow address reuse

magicrub:
When sending to a multicast port, such as 234.2.3.1:6969 for TAK, other apps on the same computer can't use that same multicast IP. Example: using WinTAK to test TAK messages from Mission Planner. There's an option in WinTAK networking settings to allow this, but for Mission Planner I think it's fine to always allow it

### DIFF
--- a/Controls/SerialOutputCoT.cs
+++ b/Controls/SerialOutputCoT.cs
@@ -171,7 +171,7 @@ namespace MissionPlanner.Controls
 
                             if (CoTStream != null && CoTStream.IsOpen)
                             {
-                                CoTStream.WriteLine(xmlStr.Replace("\r", ""));
+                                CoTStream.Write(xmlStr.Replace("\r", ""));
                             }
                         });
                     });
--- a/ExtLibs/Comms/CommsUDPSerialConnect.cs
+++ b/ExtLibs/Comms/CommsUDPSerialConnect.cs
@@ -92,7 +92,11 @@ namespace MissionPlanner.Comms
             if (IsInRange("224.0.0.0", "239.255.255.255", hostEndPoint.Address.ToString()))
             {
                 log.Info($"UdpSerialConnect bind to port {Port}");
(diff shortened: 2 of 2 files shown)


ExtLibs/Comms/CommsUDPSerialConnect.cs
```diff
@@ -92,7 +92,11 @@ public void Open(string host, string port)
             if (IsInRange("224.0.0.0", "239.255.255.255", hostEndPoint.Address.ToString()))
             {
                 log.Info($"UdpSerialConnect bind to port {Port}");
-                client = new UdpClient(int.Parse(Port), hostEndPoint.AddressFamily);
+
+                client = new UdpClient();
+                client.ExclusiveAddressUse = false;
+                client.Client.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.ReuseAddress, true);
+                client.Connect(hostEndPoint.Address, hostEndPoint.Port);
```
meee1, line 99: this actually removes the bind port, and changes to a connect vs a bind and listern

magicrub: Exactly. We don't want to bind. We want two apps to be able to use the same multicast ip and port. If you don't like that, I suppose we could add a checkbox to enable/allow port-reuse in the CoT menu

magicrub:
This is the checkbox in WinTAK to enable this
image

Controls/SerialOutputCoT.cs
```diff
@@ -171,7 +171,7 @@ void mainloop()
 
                             if (CoTStream != null && CoTStream.IsOpen)
                             {
-                                CoTStream.WriteLine(xmlStr.Replace("\r", ""));
```
magicrub, line 174: I can't remember why I added this but it doesn't really do anything and doesn't belong in this PR.

magicrub:
I'd also like to suggest having an option to ALWAYS do this, with a checkbox in the Config->Planner page.

PR #3149, opened 2023-07-18
ConfigRawParams: fix search slowdown

robertlong13:
We unwittingly introduced a search slowdown in https://github.com/ArduPilot/MissionPlanner/pull/3130. The auto column width is terribly optimized reruns any time a row's visibility is changed, and `SuspendLayout` isn't helping like I'd think it would.

To fix this, I have the auto width turn on and off. I changed the default column width mode for the entire data grid to `AllCells`, and changed the `NotSet` columns to `None`, and changed the `AllCells` columns to `NotSet`. Anywhere where `SuspendLayout` was being called, I added `Params.AutoSizeColumnsMode = DataGridViewAutoSizeColumnsMode.None;`, and restored the auto size wherever `ResumeLayout` was called (I actually implemented this in a helper function instead of copy/pasting that all 4 places).

We could also probably just go to fixed widths, but I'm happy enough with this approach now.

### DIFF
--- a/GCSViews/ConfigurationView/ConfigRawParams.Designer.cs
+++ b/GCSViews/ConfigurationView/ConfigRawParams.Designer.cs
@@ -1,4 +1,4 @@
-﻿using MissionPlanner.Controls;
+using MissionPlanner.Controls;
 
 namespace MissionPlanner.GCSViews.ConfigurationView
 {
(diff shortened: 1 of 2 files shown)


IamPete1:
Just given this a try here, Search works again, Thanks!

meee1:
i think this is why there where visible = false statement everywhere, as there is no layout performed while its not visible.

robertlong13:
> i think this is why there where visible = false statement everywhere, as there is no layout performed while its not visible.

The visible = false was only in `processToScreen`. If adding it to `filterList` fixes it, would you prefer that option to the one in this PR?

meee1:
ive made some changes.
the issue with enabled is still renders, as enabled/disabled is a valid visual state, whereas visible doesnt try render at all.

PR #3153, opened 2023-07-28
Flightdata, Payload Control tab fixes

rmackay9:
This fixes three issues I've found with the Payload Control tab

- The tab appears/disappears based on whether a MAVLink enabled gimbal has been seen or not but most gimbals are not mavlink enabled meaning this tab is normally invisible at first.  I've removed the logic to show/hide this tab and instead it is left to the user to use the Customise checkboxes.
- The Roll scroll bar direction was reversed meaning a gimbal would roll in the opposite direction to what the user would expect
- MOUNT_CONTROL messages have been replaced by MAV_CMD_MOUNT_CONTROL commands.  This command has been accepted since 4.0 (perhaps much earlier).

This has been tested on my PC with a DJI RS2 gimbal ([see AP user wiki here](https://ardupilot.org/copter/docs/common-djirs2-gimbal.html)) running latest ArduPilot.  I have not tested with older versions of AP but I think it should all work.

The only thing I was not able to test was the case where the GCS sends lat,lon,alt commands to the gimbal.  I think what I've done is correct though and from what I can see though this feature is never used in any case.

### DIFF
--- a/ExtLibs/ArduPilot/Mavlink/MAVLinkInterface.cs
+++ b/ExtLibs/ArduPilot/Mavlink/MAVLinkInterface.cs
@@ -4491,28 +4491,18 @@ Mission Planner waits for 2 valid heartbeat packets before connecting");
 
         public void setMountControl(byte sysid, byte compid, double pa, double pb, double pc, bool islatlng)
         {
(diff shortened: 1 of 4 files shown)


rmackay9:
Thanks very much!

PR #3159, opened 2023-08-11
ExtLibs: setMountConfigure uses mav-cmd-do-mount-configure

rmackay9:
This changes the SetMountConfigure method so that it uses the [MAV_CMD_DO_MOUNT_CONFIGURE mavlink command](https://github.com/ArduPilot/mavlink/blob/master/message_definitions/v1.0/common.xml#L1274) instead of the deprecated ardupilot specific [MOUNT_CONFIGURE message](https://github.com/ArduPilot/mavlink/blob/master/message_definitions/v1.0/ardupilotmega.xml#L1237).

This method is only used by the Payload Control tab's "Reset Position" button.
image

By the way, ArduPilot has consumed the do-mount-configure command since before ver 4.0 (I didn't look back any further than this).  Ardupilot stopped consuming the roll_stab, pitch_stab and yaw_stab values in version 4.3 but this PR doesn't change how these fields are set anyway.

This has been lightly tested on a real autopilot using a Xacti gimbal.  The only difference in behaviour noted was that the flight code no longer displays the deprecated message.  Below are before and after screen shots taken during testing.
mp-do-mount-configure-cmd-before-vs-after

### DIFF
--- a/ExtLibs/ArduPilot/Mavlink/MAVLinkInterface.cs
+++ b/ExtLibs/ArduPilot/Mavlink/MAVLinkInterface.cs
@@ -4469,18 +4469,15 @@ Mission Planner waits for 2 valid heartbeat packets before connecting");
         [Obsolete]
         public void setMountConfigure(MAV_MOUNT_MODE mountmode, bool stabroll, bool stabpitch, bool stabyaw)
         {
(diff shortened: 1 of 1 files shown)


rmackay9:
Thanks!

PR #3176, opened 2023-09-20
Temp: Add sysid target to safety switch command

stephendade:
Adds the target system id to the "safely switch" button.

Fixes issue when having multiple vehicles connected, pressing the "safety switch" button would not work.


robertlong13:
The safety switch can be toggled from the actions tab these days instead of this window in the latest beta. Can you confirm that's working as expected with multiple vehicles?

stephendade:
> The safety switch can be toggled from the actions tab these days instead of this window in the latest beta. Can you confirm that's working as expected with multiple vehicles?

I've now fixed that too.

The background here is Mission Planner sends out both the ``COMMAND_LONG`` and ``SET_MODE`` MAVLink messages for the safety switch. ``COMMAND_LONG`` has the correct target sysid, but ``SET_MODE`` does not.

PR #3209, opened 2023-10-14
CurrentState: add more rangefinders

robertlong13:
A partner recently asked for a 4th rangefinder on Discord. I decided to just go all the way to the ArduPilot limit like we do with batteries. Ten may be overkill, and I'm willing to pare that back if others object.


rishabsingh3003:
Thanks for this! Can we get this merged?

PR #3219, opened 2023-10-27
FlightData HUD: Fuel Gauge feature

Matthew-Alti:
New Feature:
The purpose of the fuel gauge is to display an estimation of the amount of fuel that is remaining in the tank of the vehicle. The calculation for the value displayed is based on the efi fuel consumed (in grams) and is multiplied by a ratio of the weight per litre of the fuel to get the value of the efi fuel consumed into millilitres. This calculated value is subtracted off of the fuel amount in the tank as specified by the user. This is the value which displays on the fuel gauge bar.

Functionality:
The Fuel Gauge is a new feature which has been implemented into the FlightData screen on the HUD. It is accessible by right clicking on the HUD and selecting the "Fuel Gauge" menu strip item.
image
Once the user clicks on the Fuel Gauge menu item, five controls are displayed on the persistent panel displayed under the HUD. These controls are namely: the Fuel Tank Size button, labelled as “Set Tank Size (ml)”, and input field and the Fuel Amount in the tank button, labelled as ”Set Fuel Amount (ml)”, and input field, as well as the “Set Fuel Density” button.
image
The user will first set the Fuel Density weight per litre of fuel by clicking on the set fuel density button, this will open the Fuel Density Form where there is an explanation of the use of the Fuel Density, which is the same value that the user would submit in the configuration settings in the parameter list(parameter EFI_FUEL_DENS) and that this value for the fuel gauge is only needed to be set the first time using the fuel gauge. The value is only reset if the user changes the value or the Mission Planner is reopened.
image
The user will input the weight per litre of the fuel in grams and once the Set button is pressed, the value will be used  in the fuel consumption calculation.
The user will then be able to set the size of the fuel tank in millilitres, once set the user either can press enter or press the “Set Tank Size (ml)” button which will enable the “Set Fuel Amount (ml) button and the input field for the fuel amount in the tank. Once this value is set and the user presses the button or presses enter while in the input field, the Fuel Gauge will be displayed on the HUD to the right of the Disarmed status and above where the AOA is displayed. 
image
Once the Fuel Gauge is displayed on the screen the Button’s, used for setting the fuel amount in the tank, text changes to: “Hide Fuel Gauge”. When pressed, the fuel gauge will be hidden and not display on the HUD, the button will need to be pressed again in order to display the fuel Gauge on the HUD.
image
When the amount of fuel remaining in the tank is above 30%, the gauge is displayed in a lime colour, when between 15% and 30%, the gauge is displayed in orange, when below 15%, the gauge is displayed in red.
Fuel between 15% and 30% - Orange
image
Fuel between 0% and 15% - Red
image
If a user does not want to view the Fuel Gauge bar and its controls any more, the user can right click on the HUD and select the “Fuel Gauge” menu item which will hide all of the fuel gauge controls.
image
image
When hovering over any of the controls displayed in the persistent panel for the fuel gauge input setup, the tooltip displays some text to inform the user that the fuel gauge is set and calculated in millilitres. Once the Fuel Density has been set, the Fuel Density buttons’ tooltip displays the set fuel density in grams which the user set in the fuel density form.
image
image

HUD Dropout Form
When the user double clicks on the HUD in order to view the Hud in its own form, the fuel gauge is also added onto the HUD if it was already displaying on the original HUD
image
When the HUD dropout form is resized, the fuel gauge stays in the same place relative to the size of the form.
image

The functionality works the same way when the fuel gauge is displayed on the HUD dropout as to when it is on the original HUD, the fuel gauge bar can be hidden, redisplayed, resized, the bar changes value when connected to an EFI fuel system.
Fuel Level warnings
When the fuel amount remaining reaches 15%, a message box pops up to inform the user that the fuel is below 15%, and when the fuel has reached 0, another message box displays to inform the user that the fuel has been depleted. The message boxes are only displayed when the vehicle which is connected is armed.
image
image

When a user hovers over the fuel gauge bar on the HUD, the user will be able to view the amount of millilitres remaining in the fuel tank.
image

Resizing the HUD
When the HUD is resized the fuel gauge will stay in the same position relative to the size of the HUD
image
image

### DIFF
--- a/GCSViews/FlightData.Designer.cs
+++ b/GCSViews/FlightData.Designer.cs
@@ -201,6 +201,7 @@ namespace MissionPlanner.GCSViews
             this.setHomeHereToolStripMenuItem1 = new System.Windows.Forms.ToolStripMenuItem();
             this.takeOffToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.onOffCameraOverlapToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
+            this.jumpToTagToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.but_disablejoystick = new MissionPlanner.Controls.MyButton();
             this.distanceBar1 = new MissionPlanner.Controls.DistanceBar();
             this.windDir1 = new MissionPlanner.Controls.WindDir();
@@ -225,7 +226,7 @@ namespace MissionPlanner.GCSViews
             this.scriptChecker = new System.Windows.Forms.Timer(this.components);
             this.Messagetabtimer = new System.Windows.Forms.Timer(this.components);
             this.bindingSourceStatusTab = new System.Windows.Forms.BindingSource(this.components);
-            this.jumpToTagToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
+            this.FuelGaugeToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             ((System.ComponentModel.ISupportInitialize)(this.MainH)).BeginInit();
             this.MainH.Panel1.SuspendLayout();
             this.MainH.Panel2.SuspendLayout();
@@ -323,6 +324,7 @@ namespace MissionPlanner.GCSViews
             this.hud1.batterycellcount = 4;
             this.hud1.batterylevel = 0F;
             this.hud1.batterylevel2 = 0F;
+            this.hud1.batteryon2 = true;
             this.hud1.batteryremaining = 0F;
             this.hud1.batteryremaining2 = 0F;
             this.hud1.bgimage = null;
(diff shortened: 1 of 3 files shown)


jagrandle:
This looks like a great new feature, but I think it could be refined slightly further for actual use. Why have the Set Tank Size and Set Fuel Amount boxes on two rows, using a lot of valuable vertical real estate, when they could be side by side ? On the same note, I would like to suggest having a way for these controls to be hidden, but still leaving the fuel gauge on the HUD. It seems like now you have to have the controls visible to see the gauge, or when you select hide everything disappears. Thanks

EosBandi:
It is a usefull feature, but I think most of it should be implemented on the flight controller (as script) and not in the GCS. The whole tank size/fuel density/warning part should go with the vehicle not with the gcs. It also can be combined with an average fuel consumption calculation and remaining flight time prediction.

PR #3222, opened 2023-10-31
Proximity: Fix Proximity UI concurrency issue

EosBandi:
High frequency proximity updates can cause concurrency issues on slower machines, data is updated during drawing, causing exceptions.


meee1:
My only concern here is the priority of the data vs the draw.  Ie the data shouldnt miss samples. The draw should be the weak point. If that involved a copy of the data before the draw starts that's ok

PR #3240, opened 2023-11-21
Setup: SerialPort settings screen

EosBandi:
Collects all information about serial ports on a single screen. Using rules defined in the SerialOptionRules.json to add comments and set mandatory options/baudrate when a certain protocol selected.

### DIFF
--- a/ExtLibs/Utilities/DisplayView.cs
+++ b/ExtLibs/Utilities/DisplayView.cs
@@ -95,6 +95,7 @@ namespace MissionPlanner.Utilities
         public Boolean displayParachute { get; set; }
         public Boolean displayEsp { get; set; }
         public Boolean displayAntennaTracker { get; set; }
+        public Boolean displaySerialPorts { get; set; }
 
 
         //config tuning
@@ -189,6 +190,7 @@ namespace MissionPlanner.Utilities
             displayAntennaTracker = true;
             displayRTKInject = true;
             displayJoystick = true;
+            displaySerialPorts = true;
 
 
             //config tuning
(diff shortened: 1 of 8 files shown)


rmackay9:
This sounds like a good idea, thanks!

meee1:
can i request a screenshot?

EosBandi:
Sorry, sure. 
image

rmackay9:
I'm not an official reviewer of MP PRs of course but where does the 2nd line, "UART2", etc come from?

I wonder if we could add the BRD_SER1/2_RTSCTS params alongside SERIAL1 and SERIAL2.

rmackay9:
I'd also use "baud" instead of "speed".  I know that "speed" is perhaps more understandable to the lay user but we should try and avoid synonyms I think.

rmackay9:
Sorry for the drip-feed of suggestions.. but just one more, in terms of where it appears in the "Mandatory Hardware" section, I'd probably say it should go lower in the list... perhaps just below "Servo Output".

EosBandi:
> I'm not an official reviewer of MP PRs of course but where does the 2nd line, "UART2", etc come from?
> 
> I wonder if we could add the BRD_SER1/2_RTSCTS params alongside SERIAL1 and SERIAL2? This is scope creep though, so just a suggestion.

- UART numbers are coming from the @sys/uarts.txt (if available), an idea from Andy Piper.
- BRD_SER_RTSCTS can be added; let me take a look at it... (or keep it to the next version ?)
- It can go down, no problem,
- Speed and baud are used interchangeably in the documentation with a favor for the term speed instead of baud rate.

EosBandi:
Moved tab down below Servo Outputs
Add checking and displaying BRD_SERx_RTSCTS settings
Make it a little less cluttered
image

PR #3246, opened 2023-11-29
ConfigRawParams: Various bugfixes

robertlong13:
1) Add double buffering to MyDataGridView. Prevents dangerous graphical corruption that can occur when scrolling, causing the name of the parameter you are editing to look like a different parameter. See this example; the `CAN_P1_BITRATE` parameter is falsely labeled as `CAN_LOGLEVEL`:
image
2) Fix an off-by-one error on a row-check that broke additional controls, like the bitmask button, on row 1
image
3) Stop the tree view from taking favorites into account:
image

### DIFF
--- a/Controls/MyDataGridView.cs
+++ b/Controls/MyDataGridView.cs
@@ -6,6 +6,10 @@ namespace MissionPlanner.Controls
 {
     public class MyDataGridView : DataGridView
     {
+        public MyDataGridView() : base()
+        {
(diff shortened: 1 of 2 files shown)


robertlong13:
@meee1, I'd really like to get this one in. #1 in this list can actually be pretty dangerous in the wrong circumstances (like accidentally changing a tuning value when you are trying to change a different nearby parameter).

PR #3247, opened 2023-11-30
ConfigMotorTest: support Tri frames

robertlong13:
Support additional motor test information for tri frames. Goes along with [this ArduPilot PR](https://github.com/ArduPilot/ardupilot/pull/25670), and should probably not be merged until that one is.

This brings the motor test order in line with other frames. Starts with the front-right and works around clockwise, with the addition that "Motor C" controls the tilt servo, which is a little odd, but it's exactly what "Test all in sequence" does, so we should match that.

image
image

### DIFF
--- a/APMotorLayout.json
+++ b/APMotorLayout.json
@@ -1,5 +1,5 @@
 {
-	"Version": "AP_Motors library test ver 1.1",
+	"Version": "AP_Motors library test ver 1.2",
 	"layouts": [
 		{
 			"Class": 1,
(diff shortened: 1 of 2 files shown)


IamPete1 (approved):

robertlong13:
@meee1 sorry I forgot to actually request your review once the ArduPilot partner of this PR got merged.

robertlong13:
@meee1 reminder about this one

PR #3250, opened 2023-12-01
HUD/PrearmStatus: fix failure messages

robertlong13:
When prearm checks are failing, the HUD displays the last failure message, even if that message is very stale. This changes it so it displays the latest failure message since the last time it passed the prearm check. The prearm failure messages were also moved to the lowest priority, so that they would not mask other sensor failures.

The PrearmStatus window had a similar problem, where it would show all failure messages briefly when it first launched, instead of the latest relevant ones.

Also, a drive-by fix of where the messageHigh setter function sets the _messageHighTime variable, which was causing flickers.


robertlong13:
@meee1 another simple bug fix PR when you get a chance

robertlong13:
Basically, as soon as you land, and get a prearm failure for still being in the landing sequence, you'll briefly get something like "accels unhealthy" or "IMU temperature low", which wasn't true. That was leading to some errant support questions.

robertlong13:
Here's a tlog I created so you can reproduce the behavior I'm talking about and test that the PR fixes it. Right after landing, the HUD pops up a "PreArm: Gyros inconsistent", even though the real reason it's failing is that you are in a landing sequence and are in AUTO without the next command being a takeoff.

Try replaying this log on master and with this PR to see what I mean.

[2024-09-21 08-26-07.zip](https://github.com/user-attachments/files/17081471/2024-09-21.08-26-07.zip)

Screenshot 2024-09-21 082744
Screenshot 2024-09-21 082754

The idea behind my fix is that there's no reason to assume that the current flash of not-ready-to-arm has anything to do with any previous one. This is especially true if the last case was several minutes or hours ago, but I think it's best to only display prearm failure messages that have come in since the prearm status went red.

This is a very common thing I have to explain to people
"Hey something is wrong with the aircraft, I keep getting 'gyros are inconsistent'"
"Nah, that just happens if you connect within like 10 seconds of bootup"
"No, it happened when I landed"
"No, check the messages tab, the aircraft didn't actually send that."

robertlong13:
@meee1 any thoughts on this?

PR #3258, opened 2023-12-15
CurrentState: honor MAVLinkInterface.speechenable

robertlong13:
Some speech events, like mode changes, fire off on all connected com ports, not just the primary. Additionally, `CurrentState` ignores the `speechenable` flag of the `MAVLinkInterface` that it's attached to, causing it to go off even for interfaces that are supposed to be silent.

This PR makes `CurrentState` honor the `speechenable` flag. Additionally, it sets `speechenable` on an interface when `MainV2.comPort` is set to that interface, and clears `speechenable` on the old interface that `MainV2.comPort` was pointing to.

### DIFF
--- a/ExtLibs/ArduPilot/CurrentState.cs
+++ b/ExtLibs/ArduPilot/CurrentState.cs
@@ -2827,6 +2827,7 @@ namespace MissionPlanner
                                 try
                                 {
                                     if (oldmode != mode && Speech != null && Speech.speechEnable &&
+                                        parent?.parent?.speechenabled == true &&
                                         parent?.parent?.MAV?.cs == this &&
                                         Settings.Instance.GetBoolean("speechmodeenabled") &&
                                         (armed || !Settings.Instance.GetBoolean("speech_armed_only")))
@@ -3324,6 +3325,7 @@ namespace MissionPlanner
                             try
                             {
                                 if (oldwp != wpno && Speech != null && Speech.speechEnable && parent != null &&
+                                    parent.parent.speechenabled &&
                                     parent.parent.MAV.cs == this &&
                                     Settings.Instance.GetBoolean("speechwaypointenabled") &&
                                     (armed || !Settings.Instance.GetBoolean("speech_armed_only")))


EosBandi:
What about  multi vehicle configurations ? Speech is the only feedback for changes on a non selected vehicle and this disables it as well...

robertlong13:
I was thinking about this in the context of multi vehicle applications too. "Mode changed to X", "Heading to Waypoint X", without any indication of which vehicle it came from, is, at best, unhelpful. At worst, it's overwhelming vocal spam. Especially if you have a large swarm.

If that's not convincing, then I can reduce this PR to just making CurrentState honor the speechenable flag. That way plugins have the ability to control it, and we can add options for the stock behavior in the future.

robertlong13:
@EosBandi and @meee1, thoughts on this? If you think it's best to default it so that all connections speak normally, I can reduce this PR to just fix it so that `speechenable` works correctly.

MainV2.cs
```diff
@@ -398,6 +398,11 @@ public static MAVLinkInterface comPort
             {
                 if (_comPort == value)
                     return;
+
+                // Disable speech on old port and enable it on the new one
+                if (_comPort != null) _comPort.speechenabled = false;
+                if (value != null) value.speechenabled = true;
```
meee1, line 404: what if it was disabled before hand?

robertlong13: Currently, we don't disable this on any interfaces that we also set as `MainV2.comPort`. I can't imagine a case where we want some interfaces to speak but don't want the currently selected interface to speak (if you want no interfaces to speak, this doesn't override the global `Speech.speechEnable`)

I can definitely see that the hard-coding of false on the old port, is pretty limiting though.

robertlong13:
@meee1, I have decided to limit the scope of this PR to just the bugfix. I can handle silencing/unsilencing in a plugin.

robertlong13:
@meee1 @EosBandi, any issues with this now that it's just a 2-line bugfix?

EosBandi:
I'm okay with it, but the decision to merge is ultimately up to Michael.

PR #3259, opened 2023-12-16
MAVLinkInterface: remove legacy text severity

robertlong13:
Sometimes a statustext sneaks through before the version comes in. ArduPilot 3.4 came out in 2015, so I think we can safely remove this now. I could do something backward-compatible instead if we want, but it would be messy.


EosBandi:
I think it is time to completly remove APM support and notify the user to download a time period version of MP is wants to connect to an APM system (or something that is older than 3.5)

PR #3261, opened 2023-12-21
MavlinkNumericUpDown: adjusted to consider scale for Min and Max values

tatsuy:
Implemented unit conversion for RTL_ALT's minimum and maximum values in Geo Fence screen to allow setting RTL Altitude[m] below 200m.

"RTL Altitude[m]" on the Geo Fence screen cannot be set to less than the current setting.
image

The setting must be changed in the Full Parameter List.

The cause was that the minimum value for RTL_ALT is 200 cm and the conversion of units is not taken into account.
In this PR, units are now taken into account for the minimum and maximum.

### DIFF
--- a/Controls/MavlinkNumericUpDown.cs
+++ b/Controls/MavlinkNumericUpDown.cs
@@ -84,8 +84,8 @@ namespace MissionPlanner.Controls
                     Increment = (float) Inc;
 
             _scale = Scale;
-            this.Minimum = (decimal)(Min);
-            this.Maximum = (decimal)(Max);
+            this.Minimum = (decimal)(Min / _scale);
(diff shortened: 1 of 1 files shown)


meee1:
i question this change.  as i think it will have large effects on other params

tatsuy:
Yes, it has an overall impact.
I have also confirmed the following other parameters.
PSC_VELXY_IMAX
PSC_ACCZ_IMAX
image

If this modification is not applied, it could inadvertently set an out-of-range value.
ex. set PSC_VELXY_IMAX to 4500 on Extended Tuning screen
image

It will set 4500*10 (the scale value)
image

I think it would be correct to consider scale in all parameters.

PR #3263, opened 2023-12-22
GMapMarkerBase: add options for transparent MAVs

robertlong13:
This allows the user to select whether other MAVs, or other instances of the same MAV on redundant links, should be hidden from the map, or displayed as transparent to be less distracting, or displayed as normal. The setting defaults to normal.

Additionally, added settings on the Config > Planner page to allow control over settings that were previously only available with config hacking.

Shown below is the image of one aircraft coming in over three redundant links (Radio, LTE, and SatCom) using the `Transparent` display option.
Screenshot 2023-12-22 164018

Screenshot of new config options
Screenshot 2023-12-22 163050

I had to do some funny naming nonsense for the Quad map marker. This was done keeping two things in mind:
1) Avoid constructing pens and brushes over and over assuming there is some performance penalty for that.
2) Avoid breaking plugins that might override the pen colors. Don't want to rename the existing `bluepen`, `greenpen`, or `greenbrush`.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -157,7 +157,7 @@ jobs:
       - uses: actions/upload-artifact@v2
         with:
           name: com.michaeloborne.MissionPlanner.apk.zip
-          path: ExtLibs\Xamarin\Xamarin.Android\bin\Release\*-signed.apk
+          path: ExtLibs\Xamarin\Xamarin.Android\bin\Release/${{ github.event_name == 'push' && '*-signed.apk' || '*.apk' }}
           
       - uses: "marvinpinto/action-automatic-releases@v1.2.1"
         if: ${{ github.event_name == 'push' }}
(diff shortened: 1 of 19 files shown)


meee1:
has this been tested on android build?  ie it actually runs

robertlong13:
> has this been tested on android build? ie it actually runs

I have not. I just spent a bit of time trying to, but I can't get master to run on Android right now. The CI is not actually generating the APK anymore, only an AAB. Trying to build locally, but freezing on splash (master is).

I may be doing something wrong (opening MissionPlannerLib.sln, building everything, and running the Android project). Tried on both an emulated tablet and my phone. Tried both running it through ADB and by manually installing the apk. Seems to be stuck waiting on WriteExternalStorage permission, but I only got a popup for location permission.

robertlong13:
@meee1 I had to modify CI so it could give me an unsigned APK that I could self-sign and install (I cannot, for the life of me, figure out what I'm doing wrong on local building it). It works!
Screenshot_20240515-034608

Issue #3245, opened 2023-11-28
Quick screen's climb rate should use VFR message


rmackay9:
I noticed that in the simulator the climb rate was 3x what I was expecting and it turned out that it was because I was running the simulator at 3x regular speed and MP is apparently calculating the value itself.  I think it should just use the VFR.climb value

![image](https://github.com/ArduPilot/MissionPlanner/assets/1498098/9290c578-80d4-436b-b81b-6ba696ae4925)
![image](https://github.com/ArduPilot/MissionPlanner/assets/1498098/1333ec03-e177-4b18-a12d-972f5059a01a)

PR #3267, opened 2024-01-04
CurrenState :Use VFR_HUD for climbrate IF it is received.

EosBandi:
CurrentState.climbrate will use VFR_HUD.climbrate if we receive the message. Otherwise, it falls back to the time/altitude change calculation.
This will fix: #3245


rmackay9:
Thanks for this!

Issue #3268, opened 2024-01-04
Ctrl-F Toggle Safety Switch should be two separate buttons


rmackay9:
We've had at least one user crash caused by the user "toggling" the safety switch and losing control of the vehicle when they wanted to enable movement of the servos (e.g. safety switch disabled?).

To make it safer we should replace the toggle with two buttons.
![image](https://github.com/ArduPilot/MissionPlanner/assets/1498098/d156750b-c1e6-4f37-b221-625799a1a775)

ROSStargh:
Very approving.

#2987 

#3269 
> I've experienced a crash 2 years ago when toggling safety switch under control-F. Multiple drones were connected to GCS. One was in the air, and the others were on the ground. I was trying to take off another drone.
> 
> There was no obvious indicator that shows the state of the safety switch. And the command had effect on every drone on the link. So I had no idea all drones had their safety switch already disabled.
> 
> It looks like the cause is that the command was sent to target_system=0.
> 
> ![4ec17ff6093a8556ff89373586f7628a095af31d.png](https://camo.githubusercontent.com/045a1b54f9a7dd5f7e3fe0f691e5067976a303cbb3cf9e7e192c29a8d6c662cb/68747470733a2f2f646973637573732e6172647570696c6f742e6f72672f75706c6f6164732f64656661756c742f6f726967696e616c2f33582f342f652f346563313766663630393361383535366666383933373335383666373632386130393561663331642e706e67)

PR #3269, opened 2024-01-05
prevent target_system=0 in toggle safety switch

tridge:
a user tried to toggle safety on an antenna tracker and MissionPlanner sent it to target_system=0, which resulted in toggling safety on a flying aircraft, causing it to crash
Note that we really should make it a separate on/off buttons, as per this issue: https://github.com/ArduPilot/MissionPlanner/issues/3268

This is what MissionPlanner sent:
```
2023-09-16 08:52:05.284 SET_MODE {target_system : 0, base_mode : 128, custom_mode : 1}
2023-09-16 08:52:05.308 SET_MODE {target_system : 0, base_mode : 128, custom_mode : 1}
```
I can only guess that MissionPlanner had not seen a SYS_STATUS from the antenna tracker when the button was pressed

### DIFF
--- a/GCSViews/FlightData.cs
+++ b/GCSViews/FlightData.cs
@@ -1692,8 +1692,13 @@ namespace MissionPlanner.GCSViews
                     }
                     if (CMB_action.Text == actions.Toggle_Safety_Switch.ToString())
                     {
+                        var target_system = (byte)MainV2.comPort.sysidcurrent;
+                        if (target_system == 0) {
+                            log.Info("Not toggling safety on sysid 0");
+                            return;
+                        }
                         var custom_mode = (MainV2.comPort.MAV.cs.sensors_enabled.motor_control && MainV2.comPort.MAV.cs.sensors_enabled.seen) ? 1u : 0u;
-                        var mode = new MAVLink.mavlink_set_mode_t() { custom_mode = custom_mode, target_system = (byte)MainV2.comPort.sysidcurrent };
+                        var mode = new MAVLink.mavlink_set_mode_t() { custom_mode = custom_mode, target_system = target_system };
                         MainV2.comPort.setMode(mode, MAVLink.MAV_MODE_FLAG.SAFETY_ARMED);
                         ((Control)sender).Enabled = true;
                         return;


rmackay9:
Thanks for this.  I think this protection is a good idea.
I wonder if perhaps on this system, some device was transmitting using system id 0.

EosBandi:
I see something fishy here, target_system id in the toggle command comes from the currently connected AND selected vehicle. The only way that the button send out a command with a zero sysid if the antenna tracker (or plane) sysd was set to zero, which is a serious configuration error.
```
            if (CustomMessageBox.Show("Are you sure?", "", MessageBoxButtons.YesNo) == (int)DialogResult.Yes)
                MainV2.comPort.setMode(
                    new MAVLink.mavlink_set_mode_t()
                    {
                        custom_mode = (MainV2.comPort.MAV.cs.sensors_enabled.motor_control == true && MainV2.comPort.MAV.cs.sensors_enabled.seen) ? 1u : 0u,
                        target_system = (byte)MainV2.comPort.sysidcurrent
                    }, MAVLink.MAV_MODE_FLAG.SAFETY_ARMED);
```

GCSViews/FlightData.cs
EosBandi, line 1: No need to add a variable and cast the sysidcurrent to byte
if (MainV2.comPort.sysidcurrent == 0).... is just fine
also this check must be added to temp.cs at line 1188 (CTRL-F screen).
Instead of logging, a messagebox would be better to let the user know that the button did nothing...
(CustomMessageBox.Show("Not toggling safety on sysid 0")

MichelleRos:
> Shouldn't we handle this in the AP code ?

Perhaps a good idea to put a PreArm or something to guard against that error (which may not be what happened here), but I think it's definitely a good idea to have this in here too. It's still a bug for MP to send to the broadcast id something that is only intended for a single vehicle.

ROSStargh:
I've experienced a crash 2 years ago when toggling safety switch under control-F. Multiple drones were connected to GCS. One was in the air, and the others were on the ground. I was trying to take off another drone. 

There was no obvious indicator that shows the state of the safety switch. And the command had effect on every drone on the link.

It looks like the cause is that the command was sent to target_system=0.

PR #3271, opened 2024-01-06
ConfigRawParams: When setting a parameter changes the number of parameters, do a full refresh. and check if reboot is needed

### LINKED ISSUE #3198: Param loading Green line does not go away
I think there may be an issue with the latest beta which means that the green parameter bar never goes away.

image

I can push the Config Retry button but still the bar remains.
image

EosBandi:
If a param change causing the change in the number of parameters, the parameter subsystem could stuck in a "downloading is in progress" state. This pr:
removes the name check, since not only _enable parameters can change the number of parameters
Add a check at the end of Write Params and if the reported param number not equals with the current number, it notify the user and do a param refresh.
This fixes #3198

### DIFF
--- a/GCSViews/ConfigurationView/ConfigRawParams.cs
+++ b/GCSViews/ConfigurationView/ConfigRawParams.cs
@@ -271,7 +271,7 @@ namespace MissionPlanner.GCSViews.ConfigurationView
             //}
 
             int error = 0;
-
+            bool reboot = false;
             foreach (string value in temp)
             {
                 try
@@ -283,7 +283,11 @@ namespace MissionPlanner.GCSViews.ConfigurationView
                     }
 
                     MainV2.comPort.setParam(value, (double)_changes[value]);
-
+                    //check if reboot required
+                    if (ParameterMetaDataRepository.GetParameterRebootRequired(value, MainV2.comPort.MAV.cs.firmware.ToString()))
+                    {
+                        reboot = true;
+                    }
                     try
                     {
                         // set control as well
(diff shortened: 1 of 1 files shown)


Hwurzburg:
THANKS! I see the build artifact but not sure how to run it locally for some testing....its an aab file?

davidbuzz:
param refresh dangerous when armed.

GCSViews/ConfigurationView/ConfigRawParams.cs
```diff
@@ -263,15 +263,15 @@ private void BUT_writePIDS_Click(object sender, EventArgs e)
 
             bool enable = temp.Any(a => a.EndsWith("_ENABLE"));
 
-            if (enable)
-            {
-                CustomMessageBox.Show(
-                    "You have changed an Enable parameter. You may need to do a full param refresh to show all params",
-                    "Params");
-            }
+            //if (enable)
```
peterbarker, line 266: Remove rather comment out code

rmackay9:
PeterH suggested that we should only do the full refresh when the vehicle is disarmed.

GCSViews/ConfigurationView/ConfigRawParams.cs
```diff
@@ -325,6 +329,18 @@ private void BUT_writePIDS_Click(object sender, EventArgs e)
...
+            if (reboot)
+            {
+               CustomMessageBox.Show("Reboot is required for some parameters to take effect.", "Reboot Required");
+            }
+
+            if (MainV2.comPort.MAV.param.TotalReceived != MainV2.comPort.MAV.param.TotalReported)
+            {
+                CustomMessageBox.Show("The number of available parameters changed. A full param refresh will be done to show all params.", "Params");
+                //Click on refresh button
+                BUT_rerequestparams_Click(BUT_rerequestparams, null);
```
IamPete1, line 342: Refreshing if armed might will flood the link with params, if you have a slow link this will be very painful. You could either not refresh if armed, or if armed pop up a "would you like to refresh?" message to get the user to decide.

Hwurzburg:
while testing I got this error:
image

EosBandi:
> while testing I got this error:
Based on where it happened in the code, this is unrelated. Do you have steps to repro ?


> Otherwise seems to work well....I agree with Pete's comment re refresh only while disarmed...

Indeed, added check for armed state.

PR #3277, opened 2024-01-18
Sik radio update 2024-1-18

StevenHAtRFDesign:
Updates to Mission Planner's SikRadio config screen and the SiKRadio stand-alone program from RFDTools to support most recent RFD modem features and firmware programming.


meee1:
so a few things
msi checked into git, also some firmware binary's.
failing android and osx builds

PR #3291, opened 2024-02-15
for sprung throttles the user should center the throttle

tridge:
it would be even better if there was a tick-box for sprung throttle and it automatically set the right option bit for the vehicle type. For example, on plane you would set FLIGHT_OPTIONS bit 10.


rmackay9:
In Copter the equivalent parameter is [PILOT_THR_BHV](https://ardupilot.org/copter/docs/parameters.html#pilot-thr-bhv-throttle-stick-behavior) and it's the first bit (bit0)

robertlong13:
Is it dangerous to have the throttle stick centered with unsprung sticks? Could we just set the message to "ensure all sticks are centered"?

Issue #3305, opened 2024-03-08
Plan: Loiter Radius field should be be used for multicopters


rmackay9:
The Plan screen has a "Loiter Radius" field which is probably populated from a parameter if using Plane.  Copter doesn't have this parameter though so we should hide this field and never use it (for multicopters).

This field should also NOT be used (for multicopters) when drawing the circle on the map for LOITER_TURNS commands.  If the LOITER_TURNS command's radius is zero then the circle should not be displayed because the vehicle will do a panorama shot (e.g. a pirouette, e.g. circle in place).

![image](https://github.com/ArduPilot/MissionPlanner/assets/1498098/9e5c8f63-3661-4de4-8198-9222d2d2a5db)

casrya:
@rmackay9 
I took a look at this because I use LOITER_TURNS in missions and had never noticed any problems, and found it is fact working correctly for me with Copter.

I believe it is because my value shown in the disabled loiter radius textfield (which you have highlighted in red) is set to 0.

As a workaround you can try changing it in config.xml as follows (I suspect your existing value will be 80 which you may have set using a different vehicle type?)

<TXT_loiterrad>0</TXT_loiterrad>

P.S I agree this appears to be a vehicle type bug but thought this might be helpful in the meantime?

rmackay9:
I think we should disable this field for Rovers as well as discussed on this support forum topic https://discuss.ardupilot.org/t/lidar-collision-avoidance/128497/4

Issue #3300, opened 2024-03-01
Simultaneous Display of Mission, Geofence, and Rally Points in PLAN Screen


tatsuy:
I would like to simultaneously display and edit missions, geofences, and rally points in PLAN Screen.
This would enable users to view and edit missions while also creating geofences and rally points in the context of the mission they are planning.

As it currently stands, users must toggle between "MISSION", "FENCE", and "RALLY" options via a dropdown menu located in the upper right corner of the PLAN Screen. This functionality limits the display to only one selected item at a time.

Proposal:
- Maintain the current dropdown menu for selecting the primary editable item ("MISSION", "FENCE", or "RALLY").
- The selected item from the dropdown menu would be displayed in an editable mode, allowing users to add, remove, or modify points as currently possible.
- Items not selected in the dropdown menu would be displayed in a read-only mode. This would allow users to see all elements (mission waypoints, geofence boundaries, and rally points) overlaid on the map simultaneously, but only edit the item type they have selected from the dropdown menu.

By the way, it's noted that previously, the right-click menu in the PLAN screen allowed for simultaneous creation of missions, geofences, and rally points—a feature no longer available when a vehicle is connected, as these options are removed from the menu.
https://github.com/ArduPilot/MissionPlanner/blob/latest/GCSViews/FlightPlanner.cs#L2661-L2673

tatsuy:
@EosBandi 
Thank you for your reply.
Indeed, the current functionality that allows fences to be displayed during mission editing is useful. However, it does not permit the simultaneous display of all elements—missions, fences, and rally points—on the fence or rally point screens. It would be beneficial to have the capability to edit fences and rally points while also viewing missions.

jagrandle:
I also think this is a great idea. It would be even better if it did not require the vehicle to be connected. Often more complicated missions, which are usually the ones more likely to use all these different types of points or areas, are developed in advance in the office before heading in to the field. Having to be connected to a vehicle is just another level of complexity.
Thanks

PR #3310, opened 2024-03-12
FlightPlanner: Simultaneous Display of Mission, Geofence, and Rally Points in PLAN Screen

tatsuy:
simultaneous display of all elements—missions, fences, and rally points

Changes:
- Maintain the current dropdown menu for selecting the primary editable item ("MISSION", "FENCE", or "RALLY").
- The selected item from the dropdown menu would be displayed in an editable mode, allowing users to add, remove, or modify points as currently possible.
- Items not selected in the dropdown menu would be displayed in a read-only mode. This would allow users to see all elements (mission waypoints, geofence boundaries, and rally points) overlaid on the map simultaneously, but only edit the item type they have selected from the dropdown menu.

Resolved these issues #3300 and #3305.

image

### DIFF
--- a/GCSViews/FlightPlanner.cs
+++ b/GCSViews/FlightPlanner.cs
@@ -109,6 +109,9 @@ namespace MissionPlanner.GCSViews
         private bool grid;
         private List<int> groupmarkers = new List<int>();
         private List<List<Locationwp>> history = new List<List<Locationwp>>();
+        private List<Locationwp> wpCommandList = new List<Locationwp>();
+        private List<Locationwp> fenceCommandList = new List<Locationwp>();
+        private List<Locationwp> rallyCommandList = new List<Locationwp>();
         private bool isMouseClickOffMenu;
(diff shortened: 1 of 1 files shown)


EosBandi:
It works only when you are connected to a vehicle and changes are uploaded to the vehicle.
Not working in offline mode.

tatsuy:
@EosBandi 
I've updated the implementation to support simultaneous display even when not connected to a vehicle, addressing the offline mode functionality. Feedback is always welcome!

tatsuy:
@meee1 @EosBandi 
I want to use this feature in the next lastest release.
Could someone test this for merging?

rmackay9:
Coincidentally a user just asked for this on the Rover-4.5 forums here https://discuss.ardupilot.org/t/change-fence-enable-during-mission/116690/3

LupusTheCanine:
Bump @tatsuy

PR #3312, opened 2024-03-15
GCSViews: Config DroneCAN allow with no SLCAN port

IamPete1:
Now we have MAVCAN slcan is not required, this allows the config page to still be used, only the SLCAN button is disabled. 

This allows me to use the page in SITL where the is no `CAN_SLCAN_TIMOUT`.

image


IamPete1:
I guess we could check for that one of the drivers is set to DroneCAN type, or even if there is even any CAN config parameters to disable the whole page.

PR #3313, opened 2024-03-18
mavcmd: add set-camera-source

rmackay9:
This adds the new SET_CAMERA_SOURCE mission command to the list of possible commands for copter, rover and plane.

This change requires this mavlink PR https://github.com/ArduPilot/mavlink/pull/351 to be merged which should happen at tomorrow's dev call (e.g. Mar 19th).

This has been lightly tested and seems OK to me.
image

The only issue is that we get an invalid number error because the mavlink change hasn't gone in yet.
image

### DIFF
--- a/mavcmd.xml
+++ b/mavcmd.xml
@@ -165,6 +165,15 @@
       <Y></Y>
       <Z></Z>
     </SET_CAMERA_FOCUS>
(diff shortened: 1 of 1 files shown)


rmackay9:
Thanks!

PR #3319, opened 2024-04-03
Config Planner - Choose how to access the map tiles : server, cache or both

Godeffroy:
- Add ComboBox in ConfigPlanner to choose if map is loaded from cache only, server only or cache and server (default). This setting is saved in the config.xml

- Add button in ConfigPlanner to Open Map Cache Folder

- SRTM files are downloaded if GMaps Access Mode is not CacheOnly

image

This greatly reduces cpu load and unnecessary network access in case of using an offline GCS.

I don't know if i need to modify all ConfigPlanner.resx file's language as well ?

### DIFF
--- a/ExtLibs/Utilities/srtm.cs
+++ b/ExtLibs/Utilities/srtm.cs
@@ -9,6 +9,7 @@ using System.Collections;
 using System.Net.Http;
 using System.Threading.Tasks;
 using log4net;
+using GMap.NET;
 
 namespace MissionPlanner.Utilities
 {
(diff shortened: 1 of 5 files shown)


EosBandi:
- Is there any reason why not simple check if we have network connectivity and set the mode based on the result ?  A permanent setting has a risk that user forgot it in offline mode and expects that once connected to a network it will download srtm/maptiles.
- There are already an "Age Map" function in the ctrl-f screen, if a user really want to delete the whole cache (including the database) then he/she must know where to find it :) Even if it needed, the the right place to put it is the ctrl-f screen, alongside with the other housekeeping and experimental functions.

Godeffroy:
Thanks for you response !
I plan to do another PR on the prefetch feature to make it faster, i can also add this : before prefetching, it switch whatever mode it is to ServerAndCache ?
And another PR on the inject GE as it will be possible to easily inject png, jpeg and jpg with a progress bar, change map provider to custom and map access mode to CacheOnly.

PR #3320, opened 2024-04-03
FlightPlanner : tweak prefetch, Resolves #2591, resolves #2483

### LINKED ISSUE #2483: feature request: Load offline maps
#### Issue details
Is there a feature implemented for loading offline maps? If not, I recomment to add this feature. It makes the mission planning easier if you are at the test-site without a stable wifi connection.

best regards

### LINKED ISSUE #2591: Feature Request: Enhancements to Plan > Map Tools > Prefetch
#### Enhancement details
I would like to see a couple of extra options on the Prefetch tool. Accessed via **Plan** screen, draw a range, right click, **Map Tools** > **Prefetch**.

Firstly, you can already choose the _max zoom_ level. However, an additional option to also choose the _min zoom level_ would also be great.

Secondly, and more important than the first feature, would be an option to ignore tiles that are already downloaded.

#### Version
I am using 1.3.74 build 1.3.7563.27684

#### Platform
[ ] All
[ ] AntennaTracker
[*] Copter
[*] Plane
[*] Rover
[*] Sub

Airframe type, hardware type, and logs are not applicable.

Godeffroy:
When prefetching, do not download tile if file already exist
Add minZoom input in prefetch (like maxZoom)
image

Use Parallel.For instead of For to get tiles faster !

Using original for loop to download world from zoom 5 to 1 :
image
Middle of land at zoom 12 to 10 :
image

Using Parallel.For loop to download world from zoom 5 to 1 :
image
Middle of land at zoom 12 to 10 :
image

### DIFF
--- a/ExtLibs/GMap.NET.Core/GMap.NET/GMaps.cs
+++ b/ExtLibs/GMap.NET.Core/GMap.NET/GMaps.cs
@@ -817,6 +817,31 @@ namespace GMap.NET
 
       readonly Exception noDataException = new Exception("No data in local tile cache...");
 
+      public bool CheckImageExist(GMapProvider provider, GPoint pos, int zoom, out Exception result)
+      {
+         result = null;
+         try
+         {
+            if (Mode != AccessMode.ServerOnly && !provider.BypassCache)
+            {
+               if (PrimaryCache != null)
+               {
+                  return PrimaryCache.CheckImageFromCache(provider.DbId, pos, zoom);
+               }
+
+               if (SecondaryCache != null)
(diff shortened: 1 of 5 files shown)


meee1:
there is a potential issue here
and that's with googles TOS
300 QPM (requests per minute) per IP address.

the check for existing file is all ok. its just the parallel for that's the problem

Godeffroy:
I understand the potential issue, but without modification, we can already get more than 300 tiles per minutes : 
image

meee1:
whitespace.

PR #3321, opened 2024-04-04
Remove project test to fix git action builds

EosBandi:


meee1:
Can we just remove it from the build list. Vs the project

EosBandi:
The project files were already removed from the repo a long time ago, so it is just a blind reference in the solution. Or is there any use of it ?

meee1:
i use it locally. but im flexable

PR #3323, opened 2024-04-05
Update Zoom level

Godeffroy:
FlightData : Update Zoom level
FlightPlanner : Update Zoom level
Rename trackBar1 to TRK_zoom like in FlightData
TRK_zoom take min from map provider
TRK_zoom take now value from Map zoom in float instead of int
Add Zoom Numeric Up & Down to see Zoom level

image


Godeffroy:
I could made the same layout design for the numeric up & down zoom level in Data as in Plan, just above the track bar. It's more natural to see the level as we adjust it.

Godeffroy:
Made some small arrangement in UI, also rebased on master as FlightData has changed.

Issue #2527, opened 2020-12-27
Elevation data from the Norwegian Mapping Authority.


Supermagnum:
Please implement the free and open DTM10 data from Kartverket aka the Norwegian Mapping Authority. It’s in 10x10 meter grid measurements, with a accuracy of ± 4 to 6 meters,as it’s superior to any SRTM data.

Preprocessed national dataset exports are available.
All datasets are in EUREF89, NN2000. UTM zone 33 has full coverage. Zone 32 and 35 where applicable. It is available here:
https://kartkatalog.geonorge.no/metadata/height-dtm-10/dddbb667-1303-4ac5-8640-7ec04c0e3918
There is also a API. DTM 50:
https://kartkatalog.geonorge.no/metadata/height-dtm-50/e25d0104-0858-4d06-bba8-d154514c11d2

DTM 10 API:
https://kartkatalog.geonorge.no/metadata/dtm-10-terrengmodell-utm33/dddbb667-1303-4ac5-8640-7ec04c0e3918

This looks to be 1 meters resolution:
https://kartkatalog.geonorge.no/metadata/hoeyde-dom-1-utm32-wcs/08edfa0c-0e8f-44eb-80a6-ddbf3d1b3a64

All the data is licensed under :
Creative Commons BY 4.0 (CC BY 4.0)

rmackay9:
I suspect this is similar to this issue we found for finland https://github.com/ArduPilot/ardupilot/issues/12923

Supermagnum:
> I suspect this is similar to this issue we found for finland [ArduPilot/ardupilot#12923](https://github.com/ArduPilot/ardupilot/issues/12923)

It looks like it is. I have downloaded the entire dtm data set today, it was around 6 GB .

meee1:
if you have the data as a geotiff, you can use it inside MP right now. https://discuss.ardupilot.org/t/custom-imagery-in-mission-planner/10280/13?u=michael_oborne

Supermagnum:
This looks to be 1 meters resolution:
https://kartkatalog.geonorge.no/metadata/hoeyde-dom-1-utm32-wcs/08edfa0c-0e8f-44eb-80a6-ddbf3d1b3a64

Supermagnum:
All data is licensed under Creative Commons BY 4.0 (CC BY 4.0).

Supermagnum:
[...]

PR #3338, opened 2024-04-24
srtm: removed 60N limit on SRTM downloads

tridge:
our SRTM1 and SRTM3 now cover whole world
terrain.ardupilot.org SRTM1 has covered the whole world since 2022
SRTM3 covers whole world from April 2024

### DIFF
--- a/ExtLibs/Utilities/srtm.cs
+++ b/ExtLibs/Utilities/srtm.cs
@@ -383,9 +383,6 @@ namespace MissionPlanner.Utilities
                 }
                 else // get something
                 {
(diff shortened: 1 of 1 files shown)


rmackay9:
fantastic!  I think this means we can close this AP issue. https://github.com/ArduPilot/ardupilot/issues/12923

SRTM data for above the 60N limit has appeared here http://firmware.ardupilot.org/SRTM/Eurasia/

We've also got this MP issue that can likely be closed too https://github.com/ArduPilot/MissionPlanner/issues/2527

PR #3345, opened 2024-05-03
Update Plane Config page for 4.5.x changed param names

EosBandi:
Updated params in 4.5
AIRSPEED_CRUISE, AIRSPEED_MIN, AIRSPEED_MAX (all in m/s)
and ROLL_LIMIT_DEG, PTCH_LIM_MAX_DEG, PTCH_LIM_MIN_DEGRE all in degree
Due the change in units, now params that changed are not avail in the config page with versions 4.4 and before.
It still needs some metadata update from plane to get the increments right : https://github.com/ArduPilot/ardupilot/pull/26969


tridge:
@EosBandi thanks!
image

 - the name should be "min" not "FBW mi" and "max" not "FBW ma"

the "FBW" in the name is misleading, it really has nothing to do with FBW modes, and doesn't even get used in FBWA mode.

tridge (approved):

tridge:
@meee1 ready to merge I think

PR #3365, opened 2024-06-14
Versio4.0

ganeshaero360:
File Name Change


robertlong13:
@meee1 can you please ban this account?

rmackay9:
@ganeshaero360,

Looks like you're doing lots of great stuff with Mission Planner and AP which we appreciate but maybe you could try and be more careful about which repo your PRs are raised against?

Also please remember that MP is a community driven project so please consider getting involved by contributing back to the core code base.

PR #3373, opened 2024-07-03
DashWare: fix culture bug

robertlong13:
User reported issue on Discord

image

I think this will solve it, but I'll need them to test first.


robertlong13:
The user has tested this fix and reported success.

Issue #3331, opened 2024-04-15
Log Browser: right-mouse-button menu includes odd "Properties MasterPane" option


rmackay9:
![image](https://github.com/ArduPilot/MissionPlanner/assets/1498098/8868b18d-9d3a-4df6-b05f-dc96133833bf)

The Log Browser page's graphing window has an few odd looking right-mouse-button menu items:

- Properties MasterPane
- Properties YAxis
- Properties YAxis2

These look like debug and were perhaps accidentally left in place?

rmackay9:
Sorry to be a nag but these 3 extra items are a bit annoying.  I don't think it's good that we have developer options like this exposed to the user.

EosBandi:
Only Michael knows, the commit was supposed to allow custom Y axis, but it does indeed looks like a debug menu.
I added a PR to remove it, but it is up to Michael. https://github.com/ArduPilot/MissionPlanner/pull/3376

EosBandi:
fixed in 18d17d4f936a6a4a5e85a623682cc5b4a6694b62

rmackay9:
Ah, so happy that those are gone.

PR #3376, opened 2024-07-05
LogBrowser: Remove Property pages from LogBrowser context menu

EosBandi:
As per issue https://github.com/ArduPilot/MissionPlanner/issues/3331


rmackay9:
Thank you so much!  It would be great to get rid of those 3 debug items from the right-mouse-button-click menu on the dataflash viewer screen

EosBandi:
Fixed in 18d17d4f936a6a4a5e85a623682cc5b4a6694b62

PR #3378, opened 2024-07-08
CurrentState: Add ESC13-16 telemetry

EosBandi:
the ESCs on the status page only go up to 12, but in 4.5 we support at least 16


andyp1per (approved):
Thanks!

PR #3410, opened 2024-09-17
Mcast dronecan

meee1:


rmackay9:
I wonder if maybe this has broken the ability to see DroneCAN messages from connected DroneCAN devices?  For example, I'm trying to see DroneCAN parameters from a Xacti gimbal and it doesn't show up.  Only mission planner shows up
image

robertlong13:
Just to close the loop on this, this is an ArduPilot bug https://github.com/ArduPilot/ardupilot/issues/28175

However, as I was testing this out, I did notice there's no way to disconnect the MAVLinkCAN or switch from 1 to 2 without disconnecting the autopilot. Or am I missing something?

meee1:
Did switching tabs work? Used to?

robertlong13:
> Did switching tabs work? Used to?

Ah, switching tabs works. I had only tried switching to other backstage views.

PR #3413, opened 2024-09-18
Plugin: Create Terrain DAT file for uploading

EosBandi:
You can create TERRAIN DAT files to copy to the FC's SD card from the terrain database available in Mission Planner
!Including loaded DEM files!
The resolution can be selected between 5 and 100 meters, if the available terrain data is lower resolution then the grid data will be averaged.
(less than 5 meter resolution is not possible due to a bug in Ardupilot, which prevents files larger than 2Gb on SD cards)

Use :
Go to FlightPlanner.
Select area with ALT+MouseCLICK and drag
Open context menu and select Make Terrain DAT

if no area is selected the actual displayed area is used.

Note: DAT files are always created for one by one degree area, the size depends on the resolution.

### DIFF
--- a/MissionPlanner.sln
+++ b/MissionPlanner.sln
@@ -3,14 +3,14 @@ Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio Version 17
 VisualStudioVersion = 17.3.32811.315
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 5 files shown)


meee1:
I like it. I often thought of this myself.  Side question. Do you have a parser side as well?

EosBandi:
>  Side question. Do you have a parser side as well?
I did not see a use case for parser, so I skipped it.

PR #3423, opened 2024-09-28
Add MAVLink forwarding via WebSocket server

IamPete1:
This adds forwarding to WebSockets. I have only tested sending from MissionPlanner to the WebSocket not the other way (my application does not support send). So far I have been using pymavlink with https://github.com/ArduPilot/pymavlink/pull/967.

This very hacky currently. It doesn't work if you open the same connection more than once for example. I'm not cleaning up correctly I guess, but this is right on the edge of my c# knowledge. 

This does add Fleck as a new dependency, it makes the server stuff easier.  https://github.com/statianzo/Fleck

This is more or less a Fleck example and code copied from the existing client WebServer stuff.

### DIFF
--- a/Controls/SerialOutputPass.Designer.cs
+++ b/Controls/SerialOutputPass.Designer.cs
@@ -100,7 +100,8 @@
             this.Type.Items.AddRange(new object[] {
             "Serial",
             "TCP",
-            "UDP"});
+            "UDP",
+            "WS"});
             this.Type.Name = "Type";
             this.Type.Resizable = System.Windows.Forms.DataGridViewTriState.True;
             this.Type.SortMode = System.Windows.Forms.DataGridViewColumnSortMode.Automatic;
--- a/Controls/SerialOutputPass.cs
+++ b/Controls/SerialOutputPass.cs
@@ -241,6 +241,17 @@ namespace MissionPlanner.Controls
                         MainV2.comPort.MirrorStream.BaudRate = int.Parse(extra);
                         MainV2.comPort.MirrorStream.Open();
                     }
+                    else if (protocol == "WS")
+                    {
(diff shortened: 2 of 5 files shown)


meee1:
https://github.com/ArduPilot/MissionPlanner/blob/693c6e3445af310d90fe90daa29f40df6bd64fc6/Utilities/httpserver.cs#L180

This is the internal server I mentioned. 

Looks like your code is a client though, vs the existing is serverside

meee1:
My use case was a browser connecting to MP as the data source

Yours is to push data to somewhere?

meee1:
https://github.com/ArduPilot/MissionPlanner/blob/693c6e3445af310d90fe90daa29f40df6bd64fc6/ExtLibs/Comms/CommsWebSocket.cs#L16

This one is for connecting to a socketio ws service as a client

IamPete1:
> My use case was a browser connecting to MP as the data source
> 
> Yours is to push data to somewhere?

Right, mission planner connects directly to the vehicle then it forwards onto a web based dashboard.

meee1:
I assume you have something working? As I would have thought the dashboard makes the connection to the backend vs the backend connecting to the frontend 

Placeholder 
Ws://127.0.0.1:56789/websocket/server

IamPete1:
> I assume you have something working?

https://github.com/IamPete1/WebTools/tree/MAVLink_Dashboard

As with all the WebTools it is 100% client side, so the only option we have is WebSocket client (we could do direct COM ports too but I haven't implemented).

IamPete1:
Turns out MissionPlanner already forwards on `ws://127.0.0.1:56781`. So we don't need this PR. https://github.com/ArduPilot/WebTools/pull/196 adds auto connection.

PR #3429, opened 2024-10-15
ADSB: uAvionix Transponder Updates

nicholas-inocencio:
This PR contains updates to the interface with the ping200X in the Transponder tab. Co-requisite with the ArduPilot PR https://github.com/ArduPilot/ardupilot/pull/28419

Changes:

- Fix a display error where the "ON" button was not properly bolded when the transponder mode was set to "ON"
  - It should be bold when the status message indicates mode A, S, and 1090ES are on and mode C is off
- Requests the UAVIONIX_OUT_STATUS mavlink message at 1 Hz when it first gets that same message
  - This enables smarter timeout behavior; see next point
- Make controls/display more responsive to connection errors with the transponder
  - Display "Transponder Offline" when status message is received but the flag `UAVIONIX_ADSB_OUT_STATUS_FAULT_STATUS_MESSAGE_UNAVAIL` is set.
  - Display "Transponder Status Lost" if the GCS had previously received a UAVIONIX_OUT_STATUS message but 5 seconds pass without seeing another one. 
    - This usually happens if the AP is power cycled (thus resetting the UAVIONIX_OUT_STATUS request)

### DIFF
--- a/GCSViews/FlightData.cs
+++ b/GCSViews/FlightData.cs
@@ -6252,8 +6252,8 @@ namespace MissionPlanner.GCSViews
                                                              !Mode_clb.GetItemChecked(2) &&
                                                              !Mode_clb.GetItemChecked(3)) ? FontStyle.Bold : FontStyle.Regular);
                     ON_btn.Font   = new Font(ON_btn.Font,   ( Mode_clb.GetItemChecked(0) &&
-                                                              Mode_clb.GetItemChecked(1) &&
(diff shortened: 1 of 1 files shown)


magicrub:
I just tested this along with the ArduPilot changes and it looks like a good improvement. I'm retesting on Friday and when that merges then we should merge this too

magicrub:
I've tested this while using PR https://github.com/ArduPilot/ardupilot/pull/28419 and it works much better. Let's merge!

PR #3431, opened 2024-10-16
paramcompare: remove double inequality check

robertlong13:
Fixes bug reported by Alex/ARACE on Discord:
image

There might be a more efficient way to do a double compare that accounts for epsilon differences, but this works.


EosBandi (approved):
Indeed it is quick fix, but it is working.

PR #3433, opened 2024-10-21
Add interface for plugins to add HW and SW config pages

EosBandi:
It makes possible for a plugin to add pages to the SETUP and the CONFIG pages.

 Simply create a MyUserControl based form and then add it via
` SoftwareConfig.AddPluginViewPage(typeof(pageclass), "pagename");`
or 
` InitialSetup.AddPluginViewPage(typeof(pageclass, "pagename");`

### DIFF
--- a/GCSViews/InitialSetup.cs
+++ b/GCSViews/InitialSetup.cs
@@ -6,6 +6,7 @@ using MissionPlanner.GCSViews.ConfigurationView;
 using MissionPlanner.Radio;
 using MissionPlanner.Utilities;
 using System;
+using System.Collections.Generic;
 using System.Reflection;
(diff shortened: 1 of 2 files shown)


robertlong13 (approved):
This is great. Been wanting something like this for a while.

robertlong13:
Do you have a minimal example plugin for this? I could whip one up if you don't.

EosBandi:
How we shall distribute these examples ? It won't work as a single file .cs since it needs a form...

meee1:
Check generator.cs. Definitely doable

EosBandi:
:) Ok put together a small example.

PR #3435, opened 2024-10-29
ConfigSerialInjectGPS.cs: Fix Septentrio config error while connected to ublox and changing away from the config RTK/GPS screen.

EosBandi:
Fix Septentrio config error while connected to ublox and changing away from the config RTK/GPS screen.
if "SerialInjectGPS_SeptentrioRTCMLevel" is not set, then the combo box is set to 1, which triggers the update....


yuri-rage:
Looks like a simple logic error and equally simple fix!

PR #3438, opened 2024-11-05
Example22: Payload Select Config Page

robertlong13:
Adds an example for the new plugin config/setup page interface: #3433.

This example implements a payload selection page to reconfigure parameters based on what payload you have installed:
image

By default, this plugin doesn't load. Someone has to add payloads to the .cs file first.

If we like this, I'll write up a post about it on Discuss.


meee1 (approved):

EosBandi (approved):
It is a better example than what I had in my mind, since it does have a valid function, not just a simple example.
Thanks for adding it !

PR #3451, opened 2024-11-24
FlightData: Actions: add engine start/stop

robertlong13:
Allows IC Engines to be used without an RC switch.

### DIFF
--- a/ExtLibs/ArduPilot/Mavlink/MAVLinkInterface.cs
+++ b/ExtLibs/ArduPilot/Mavlink/MAVLinkInterface.cs
@@ -2501,6 +2501,19 @@ Mission Planner waits for 2 valid heartbeat packets before connecting
         }
 
         /// <summary>
+        /// Control ICE Engine
+        /// </summary>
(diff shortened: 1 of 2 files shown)


ExtLibs/ArduPilot/Mavlink/MAVLinkInterface.cs
```diff
@@ -2500,6 +2500,19 @@ public bool doHighLatency(bool onoff= false)
             return ans1;
         }
 
+        /// <summary>
+        /// Control ICE Engine
+        /// </summary>
+        /// <param name="onoff">true to enable, false to disable</param>
+        /// <returns></returns>
+        public bool doEngineControl(bool onoff = false)
```
meee1, line 2508: can we change to sysid and compid needed here.

im trying to make this class able to target any connected mav. feed the sysid and compid from the ui level

robertlong13: Good call, sorry.

Done.

PR #3452, opened 2024-11-25
Fix ADSB Squawk back to decimal

magicrub:
This [commit](https://github.com/ArduPilot/MissionPlanner/commit/eefd2167358c5c121c61b58f0eab4f85034d5ef8#diff-14ff4d1be768ca629959b5d1f5efdb8058bd7b9c448b74681ef87c509eac4c89R4074) by @MUSTARDTIGERFPV changed the squawk text to be shown in Hex. I'm not sure why, maybe he knows? I've never seen it expressed as a decimal. Technically it's an octal number but decimal is the normal conversion.

### DIFF
--- a/GCSViews/FlightData.cs
+++ b/GCSViews/FlightData.cs
@@ -4083,7 +4083,7 @@ namespace MissionPlanner.GCSViews
 
                                     adsbplane.ToolTipText = "ICAO: " + pllau.Tag + "\n" +
                                                             "Callsign: " + pllau.CallSign + "\n" +
-                                                            "Squawk: " + pllau.Squawk.ToString("X4") + "\n" +
+                                                            "Squawk: " + pllau.Squawk.ToString("0") + "\n" +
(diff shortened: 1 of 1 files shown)


MUSTARDTIGERFPV:
Hey @magicrub! Unfortunately this change doesn't match normal aeronautical standards. A quick check shows squawk values with 5 digit decimal representations. When squawk values are shown for display in applications, they're actually a representation of the octal value as a 4-digit hex value, *not* as a decimal number.

Take as an example, N635ND flying over the Seattle area right now. They're squawking **1200**, the standard US squawk code for VFR flight. Here's that displayed by tar1090.
image

This matches how a squawk value is entered into a transponder and read over the radio.
image

By comparison, with your change, now Mission Planner shows this squawk code as **4608**. Excuse the weird font, I'm running on macOS because that's what I have available at the moment.
image

This PR should be reverted.

PR #3455, opened 2024-11-30
Travelled Distance Calculation Fix (Update CurrentState.cs)

Mirghani094:
Hello, this is my first pull request for Mission Planner or GitHub in general. Hence, I hope I am doing this correctly.
 
I noticed some of the Current State (MAV.cs) parameters are calculated inside Mission Planner (Travelled Distance and Time in Air), which sometimes leads to mistakes in calculations (for example, if the link got disconnected for sometime or the Mission Planner was restarted while UAV is flying).
Hopefully we can find a solution to solve this issue.

For now, I found a bug when **changing the Distance Unit while the UAV is flying**, the travelled distance calculation goes wrong. I found that the variable "disttravelled" is accumulated from previous value regardless of the distance unit (meters or feet).

I Updated CurrentState.cs  as shown in my commit, perhaps there is a better way that you can think of.

### DIFF
--- a/ExtLibs/ArduPilot/CurrentState.cs
+++ b/ExtLibs/ArduPilot/CurrentState.cs
@@ -1139,7 +1139,8 @@ namespace MissionPlanner
 
         [DisplayFieldName("distTraveled.Field")]
         [DisplayText("Dist Traveled (dist)")][GroupText("Position")] public float distTraveled { get; set; }
-
+  	public string distTravelledUnit = CurrentState.DistanceUnit;  // added by Mir 30112024
+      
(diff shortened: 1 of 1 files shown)


jagrandle:
This is a very interesting point. We have had the same sort of issues with the **Time in Air** value. It seems to be calculated by the GCS and is only incremented as long as the aircraft is connected. This means that during pauses or breaks in connection it does not increment, and thus later displays a value that is too low.

This may be the wrong place for the request as it is not strictly solely Mission Planner, though changes to MP may be required, but can we please consider having  **Time in Air**  measured or calculated on the aircraft and just reported to Mission Planner ?
Thank you.

PR #3459, opened 2024-12-13
SITL: add version selector dropdown

### LINKED ISSUE #3457: added Beta to SITL download options
For beta testing it would be very helpful if users could test the beta in MissionPlanner SITL
Currently it gives this:
image
I've added the Beta directory here:
https://firmware.ardupilot.org/Tools/MissionPlanner/sitl/Beta/
can we make the fw selection be between Stable, Beta and Latest(dev) ?

robertlong13:
Replaces the "Skip Download" checkbox with a version selection dropdown (which also contains a Skip Download option). The selected version persists between launches.

This also enables Linux users to select their version too (I believe). I don't have a good way to test this though.

Also adds the new beta download URL (Closes #3457)


robertlong13:
image

PR #3464, opened 2024-12-24
FlightData: update CMB_mountmode population

### LINKED ISSUE #3462: Data, Action Mount mode blank
For some reason the Data screen's Action Tab's Mount drop-down appears to be blank.  I'm not sure how MP populates this drop-down but if it's from a flight code parameter I'm happy to fix asap.  In any case, it's quite important that users be able to change the mount mode

image

robertlong13:
Fixes #3462

Also adds two modes that were missing because we were looking for the old parameter name.


casrya:
FYI there is also a mount mode combo in class Joy_Mount_Mode for the joystick settings.

PR #3468, opened 2024-12-27
SerialSupportProxy: add TCP option

### LINKED ISSUE #3465: support proxy TCP
The support proxy at support.ardupilot.org now supports both TCP and UDP. It would be good to add a drop down in the connection box allowing users to select UDP or TCP. The big advantage of TCP is no packet loss, allowing for reliable param download using FTP and DroneCAN peripheral management even on poor links.
The downside of TCP is you can end up with a large TCP backoff, so UDP is recommended for support sessions where the vehicle may be flying.

robertlong13:
Resolves #3465


robertlong13:
image

I don't have any warnings or guidance for which to pick, but I think it's safe to assume the support engineer will know which is appropriate and guide the user.

tridge (approved):
fantastic, thank you!

PR #3469, opened 2024-12-29
Fix ESP32-C3 Dronebridge Reset on First connect on Windows

EosBandi:
For some weird reason the WinSerial driver coupled with the ESP32 SerialOverUSB drive changes the RTS line at connect. But only at the first connect after power up.
The fix is to set RTS/CTS handshake enable but just for the open command, and disable it immediately.
I added a new option to Planner Settings to enable this function, so it can be enabled only on when needed, and does not interfere with any other serial settings.

Note: If you once had a reset on connect, it stuck so you have to remove/reconnect the esp32 before trying again.

### DIFF
--- a/ExtLibs/Comms/CommsSerialPort.cs
+++ b/ExtLibs/Comms/CommsSerialPort.cs
@@ -145,6 +145,10 @@ namespace MissionPlanner.Comms
 
         public void Open()
         {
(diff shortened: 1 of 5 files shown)


rmackay9:
This is great to see, thanks so much for this!

rmackay9:
@meee1 Thank you so much for merging!

PR #3473, opened 2025-01-06
add XC-ESC Technology DroneCAN ESC Support!

lcago:


meee1:
please add your new dsdl's to here
https://github.com/dronecan/DSDL/

PR #3475, opened 2025-01-08
Add unitType attribute to mavcmd.xml

robertlong13:
This fixes an issue where commands that used the <Z> parameter for anything other than altitude were getting unit conversions applied when they shouldn't have been. I'm also adding a unit label to the altitude column.

This also opens up the possibility of converting units (distance or altitude) for other columns, which I'll do next, probably as part of this PR, but I haven't gotten that far yet.

This will take some careful review to make sure I haven't screwed things up. I looked long and hard at all the places in FlightData that look at `multiplieralt`, but I could have missed something. I've done some basic testing, and it appears to be working.

One thing I noticed is that we don't seem to updata `mavcmd.xml` when we update Planner. If the file exists, we don't overwrite it with the one baked into the resources, which seems to be a problem IMO. Someone running Planner that was installed ages ago might not have up-to-date mavcmds, unless I'm mistaken. I can understand not wanting to immediately clobber a customized file, but we have other paths to add custom mission items in config.xml. I made sure to try to handle legacy files just in case.

### DIFF
--- a/ExtLibs/Xamarin/Xamarin/Resources/mavcmd.xml
+++ b/ExtLibs/Xamarin/Xamarin/Resources/mavcmd.xml
@@ -10,44 +10,17 @@
       <P4></P4>
       <X>Lat</X>
       <Y>Long</Y>
-      <Z>Alt</Z>
+      <Z unitType="alt">Alt</Z>
     </WAYPOINT>
-    <SPLINE_WAYPOINT>
-      <P1>Delay</P1>
-      <P2></P2>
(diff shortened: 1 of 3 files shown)


robertlong13:
Also, I noticed the Xamarin resources mavcmd file was way out of sync with the other one. Was this intentional? The Xamarin file is still the one used by Android, iOS, etc, right? Or is it defunct? I added a commit syncing that, but I'm not 100% sure if that's right.

PR #3476, opened 2025-01-08
Fix MavLink Connection Timeout

ArthurPatriot:
If `CONNECT_TIMEOUT_SECONDS` greater than 60 seconds remain seconds calculated in wrong way. And we can get maximum 60 seconds timeout.


robertlong13 (approved):
Yup, classic mistake. Great catch.

PR #3477, opened 2025-01-13
Correct Plane Loiter2alt mission command params

Hwurzburg:


tom-sovps:
@Hwurzburg @meee1 are we sure this was intended behaviour? Checking [Ardupilot](https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Mission/AP_Mission.cpp) and [mavlink](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LOITER_TO_ALT) source indicates that param2 is the radius/direction and param4 is xtrack?

robertlong13:
My response was too aggressive. I was very frustrated seeing that this slipped my notice.

The P4 term wasn't wrong. It's not perfect, but I think "exit tangent" is an acceptable summary of "track along the tangent line after exiting". Even so, simply removing it didn't make sense.

This PR had a reviewer, but as you point out, the review was probably just checking the wiki, which was changed without review. I'm not saying that we need to add more review to the wiki in general (which I agree would be untenable, and you do a lot of great work there), but if I found a **huge** discrepancy between the docs and ArduPilot (the radius being entirely in the wrong field for 10 years would be a huge deal), I would be putting a lot of explanation into both PRs, explaining why I believed it was wrong, when/why it went wrong, and invite conversation on it (in the PR, or on Discord, or on the dev call, something) to make sure I wasn't crazy. This PR and the wiki PR are both entirely wordless.

PR #3481, opened 2025-01-21
Fix popout StartupLocation bug

robertlong13:
There is a bug where a user can be persistently stuck with a full-screen sized, or a zero sized, EKF status window (or Vibe or PreArm messages windows). We don't allow the user to resize these windows, so the only way to fix it is with config.xml hacking (or deleting).

We hide the minimize/maximize buttons on these windows, but they can still be minimized or maximized in multiple ways: keyboard shortcut, double-clicking the titlebar, or from the context menu on the title bar. If the user does this, and then closes the window before restoring it, the size gets saved and it cannot be fixed through the UI.

This PR fixes this by saving the location and size of the `RestoreBounds` of the form, instead of the literal size.

Additionally I made the EKF/Vibe/PreArm windows always open in normal state, regardless of the state they were in when they closed.

### DIFF
--- a/ExtLibs/Controls/ControlHelpers.cs
+++ b/ExtLibs/Controls/ControlHelpers.cs
@@ -45,7 +45,7 @@ namespace MissionPlanner.Controls
         /// <param name="control">Control to save the settings for</param>
         public static void SaveStartupLocation(this Form control)
         {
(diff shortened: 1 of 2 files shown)


rmackay9:
It's great to see this!  A company here in Japan hit this issue and I had to get them to delete their config to restore their MP to working order

PR #3483, opened 2025-02-01
MAVLinkInterface.cs: remove 0.9 from mavlink parser

peterbarker:
treating 'U' as a header marker byte leads to the problem that we will accidentally consume parts of a valid packet after receiving a "plain text" message.

This is particularly problematic in ArduPilot's config-error-loop, as we emit the problem as both statustext and raw "line noise" to the GCS; the statustext is lost, mneaning the user doesn't get any specific error message from the autopilot, just "in config error loop".


robertlong13 (approved):
Tested using https://github.com/ArduPilot/ardupilot/pull/29207 before and after this patch, and it works as expected.

Screenshot 2025-02-01 235841

PR #3484, opened 2025-02-01
MAVLinkInterface: tidy plaintxtline printing

robertlong13:
Very minor nitpick I noticed when I was testing #3483. This was putting the newline in the wrong spot in the console.


robertlong13:
before
after

PR #3502, opened 2025-03-12
Populate VFR_HUD heading into MP HUD yaw display

Ryanf55:
# Purpose

Allow the yaw field from [VFR_HUD](https://mavlink.io/en/messages/common.html#VFR_HUD) to populate the mission planner yaw field.

According to the message docs:
```
Metrics typically displayed on a HUD for fixed wing aircraft.
```

Given that, I think the yaw field should be displayed.

# Demo

Using MAVROS, I publish mavros VFR_HUD, which gets converted into MAVLink VFR_HUD.
 
 ```
 ros2 topic pub /mavros/vfr_hud mavros_msgs/msg/VfrHud "{header: auto, altitude: 5.0, climb: 6.0, airspeed: 1.28, groundspeed: 7.0, throttle: 8.0, heading: 9.0}
 ```
 
 Then, you can see mission planner show the data correctly.
image

Before this change, VFR_HUD's heading was not parsed.

### DIFF
--- a/ExtLibs/ArduPilot/CurrentState.cs
+++ b/ExtLibs/ArduPilot/CurrentState.cs
@@ -3760,6 +3760,7 @@ namespace MissionPlanner
                             groundspeed = vfr.groundspeed;
                             airspeed = vfr.airspeed;
                             ch3percent = vfr.throttle;
+                            yaw = vfr.heading;
 
                             if (sensors_present.revthrottle && sensors_enabled.revthrottle && sensors_health.revthrottle)
                                 if (ch3percent > 0)


meee1:
Yaw and heading are not the same thing. That's why this link doesn't exist currently

Ryanf55:
Ah, yea, looks like there's a distinction in ArduPilot for these two. Seems like we should use ATTITUDE for the GCS to report yaw on the HUD.

For anyone curious:

https://github.com/ArduPilot/ardupilot/blob/d64cf3ec2d35be4ee86cac0d8fcd9adf4baeacc6/libraries/GCS_MAVLink/GCS_Common.cpp#L3388C9-L3388C28

https://github.com/ArduPilot/ardupilot/blob/d64cf3ec2d35be4ee86cac0d8fcd9adf4baeacc6/libraries/GCS_MAVLink/GCS_Common.cpp#L5796

Which are populated by the two places:

https://github.com/ArduPilot/ardupilot/blob/d64cf3ec2d35be4ee86cac0d8fcd9adf4baeacc6/libraries/AP_AHRS/AP_AHRS.h#L561


https://github.com/ArduPilot/ardupilot/blob/d64cf3ec2d35be4ee86cac0d8fcd9adf4baeacc6/libraries/AP_AHRS/AP_AHRS.h#L586

PR #3505, opened 2025-03-16
Add white outline to wpno on wp markers for better visibility.

EosBandi:
image


robertlong13:
I agree with Michael on point number 1

PR #3538, opened 2025-06-02
Fix quaternion gimbal lock

vr4guid:
fix _dcm_to_euler calculation when pitch is -pi/2


meee1:
all my sources indicate it was correct before this change.

robertlong13:
> This corrected the issue on our side.

How? Isn't this a non-functional change? `phi` is set to 0 in the previous line. Now, that raises the question of what the `- phi` and `+ phi` were ever doing in these two places to begin with, but this change can't possibly have done anything, unless I'm gravely mistaken (and I might well be; I keep reading it over and over).

PR #3544, opened 2025-06-10
Codex/перевести описания на русский язык

VermennikovMV:


yuri-rage:
Recommend closing without merge, as this affects all users, and we already have language compatibility settings available in CONFIG/Planner.

EosBandi:
This has nothing to do with the public Mission Planner, it is about creating russian version. With many hardcoded translations....

PR #3552, opened 2025-06-18
ODID: send 0/0 operator location

tridge:
if the operator GPS becomes unplugged we should tell the flight controller that we no longer have a position, so pre-arms can fail correctly

### DIFF
--- a/Plugins/OpenDroneID2/OpenDroneID_UI.cs
+++ b/Plugins/OpenDroneID2/OpenDroneID_UI.cs
@@ -297,30 +297,22 @@ namespace MissionPlanner.Controls
             try 
             {
                 // Check NMEA GPS information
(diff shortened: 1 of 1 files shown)


Plugins/OpenDroneID2/OpenDroneID_UI.cs
```diff
@@ -297,30 +297,22 @@ private void checkGCSGPS()
             try 
             {
                 // Check NMEA GPS information
-                NMEA_GPS_Connection.PointNMEA _gps_data = nmea_GPS_1.getPointNMEA();
+		NMEA_GPS_Connection.PointNMEA _gps_data = nmea_GPS_1.getPointNMEA();
```
robertlong13, line 300: Your editor seems to have replaced the spaces with tabs

PR #3555, opened 2025-06-26
Update efi fuel flow units

lbullard-c2:
The mavlink efi status message uses units of cm3/min, not g/min as displayed currently in MP


robertlong13 (approved):
This is correct, so I approve.

I noticed this a long time ago, but I always had my aircraft set up to lie and send grams in that message, so this error played to my benefit.

PR #3560, opened 2025-07-14
FlightPlanner: Add units to Takeoff prompt & skip pitch for VTOL

LachlanConn:
Updates the takeoff altitude prompt to display the current altitude unit.
The default takeoff altitude is adjusted based on selected unit.
Will Skip the Pitch prompt for Quadplanes if Q_OPTIONS bit 1 is disabled
Relies on reading params so does rely on connection to a quadplane to skip pitch
May need to add a setting in planner at some point to have a default firmware type when disconnected from an aircraft.


pompecukor:
@EosBandi are you a maintainer? Can you review please?

pompecukor:
@meee1 please if you have time.

robertlong13 (approved):
Tested and reviewed. LGTM

PR #3561, opened 2025-07-18
Update OpenSeaMapHybridProvider.cs

EosBandi:
Prevent compiler optimization that stops OpenSeaMap to  work.


meee1:
not sure on how this would be optimized out.

PR #3568, opened 2025-07-27
Scripts: Modernize Python 2 code to get ready for Python 3

cclauss:
% `ruff check Scripts | grep SyntaxError`
% `uvx --python=3.12 --from=future futurize --stage1 --write <files with SyntaxErrors>`

https://docs.astral.sh/ruff
https://github.com/PythonCharmers/python-future


tpwrules:
Do note that Mission Planner uses IronPython rather than CPython. Python 3 is still a very new thing for that AFAIK.

PR #3574, opened 2025-08-17
FlightPlanner.cs: Warn operator about zero altitude in a WP

EosBandi:
It seems that it is still not common knowledge that 0 alt in WP means no altitude change. I got some crash reports from users who generate missions in other platforms and use them from Mission Planner.


GCSViews/FlightPlanner.cs
```diff
@@ -705,6 +705,18 @@ public void BUT_write_Click(object sender, EventArgs e)
                         }
                     }
                 }
+                ushort cmdtochk = getCmdID(Commands.Rows[a].Cells[Command.Index].Value.ToString());
+                //Check for zero alttitude
+                if (cmdtochk == (ushort)MAVLink.MAV_CMD.WAYPOINT &&
+                    double.Parse(Commands[Alt.Index, a].Value.ToString()) == 0)
+                {
+                    if (DialogResult.OK != Common.MessageShowAgain("Zero Altitude Warning",
+                        "WP#" + (a + 1) + " has zero altitude, this means no altitude change! Do you want to continue or cancel wp upload ?",
```
robertlong13: I'd add a hint that if they really want 0 altitude, they should write a small value like 0.01.

GCSViews/FlightPlanner.cs
```diff
@@ -1955,6 +1967,18 @@ public void but_writewpfast_Click(object sender, EventArgs e)
                         }
                     }
                 }
+                ushort cmdtochk = getCmdID(Commands.Rows[a].Cells[Command.Index].Value.ToString());
+                //Check for zero alttitude
+                if (cmdtochk == (ushort)MAVLink.MAV_CMD.WAYPOINT &&
+                    double.Parse(Commands[Alt.Index, a].Value.ToString()) == 0)
+                {
+                    if (DialogResult.OK != Common.MessageShowAgain("Zero Altitude Warning",
```
robertlong13: Minor nit: I'd break the warning out into a function so they warning messages don't get out of sync if someone changes them and doesn't realize there's another one.

GCSViews/FlightPlanner.cs
EosBandi: Done.

GCSViews/FlightPlanner.cs
EosBandi: Also done

PR #3582, opened 2025-09-03
My button : add default value to TextColorNotEnabled

Godeffroy:
When creating a custom button based on MyButton, there is unecessary code for TextColorNotEnabled.
By adding the defaut value into the base code, it's not longer generated.
Remove this properties when the default value is used in every files.

Remove properties BGGradBot BGGradTop Outline TextColor because it's the default values.


Godeffroy (commented):
Update with master branch

PR #3591, opened 2025-09-25
FlightPlanner: fix zero-alt warning for ArduPlane

robertlong13:
Draft PR. I've run out of time today. I'll try to come back to this next week or Andras can maybe finish it; just want to get what I have out there.

So, on Copter, pretty much any nav command with 0 altitude means "keep current altitude". On Plane, this is only true for loiters (and only relative altitude). Waypoints in plane will happily use 0 as a true 0. I haven't tested across different releases yet.

I only just realized this today. The recent addition trying to inform users about the nuances of 0-alt commands (#3574) is actually dangerously wrong for plane users.

We probably need an emergency release after this.

TODO:
- The showagain form doesn't play nice with long messages. I need to try to fix that without changing the old warnings too much: 
image

- I only partially applied the rename of the existing warning variable. The designer didn't apply the rename in the other languages.
- The new warning needs the german translation

### DIFF
--- a/ExtLibs/Strings/Strings.Designer.cs
+++ b/ExtLibs/Strings/Strings.Designer.cs
@@ -1700,11 +1700,24 @@ namespace MissionPlanner {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to WP# {0} has zero altitude, this means no altitude change! If you want zero altitude, change it to a 0.01m. Do you want to continue or cancel wp upload ?.
+        ///   Looks up a localized string similar to WP# {0} has zero altitude, this means no altitude change! If you want zero altitude, change it to 0.01. Do you want to continue or cancel wp upload?.
         /// </summary>
-        public static string ZeroAltWarning {
+        public static string ZeroAltWarningCopter {
             get {
-                return ResourceManager.GetString("ZeroAltWarning", resourceCulture);
+                return ResourceManager.GetString("ZeroAltWarningCopter", resourceCulture);
+            }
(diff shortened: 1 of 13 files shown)


robertlong13:
@meee1 the CI failure seems to be somehow unrelated to this PR (a recent change on the default runner maybe?). I just ran it on master on my fork and [got the same error](https://github.com/robertlong13/MissionPlanner/actions/runs/18035620596):

```
Xamarin.Android.Tooling.targets(100,5): error XA5207: Could not find android.jar for API level 33. This means the Android SDK platform for API level 33 is not installed. Either install it in the Android SDK Manager (Tools > Android > Android SDK Manager...), or change the Xamarin.Android project to target an API version that is installed. (C:\Program Files (x86)\Android\android-sdk\platforms\android-33\android.jar missing.) [D:\a\MissionPlanner\MissionPlanner\ExtLibs\Xamarin\GDAL\GDALMPAndroid.csproj]
```

robertlong13:
Discussed this on the dev call, and Tridge has signed off on the message and logic (for plane).

meee1:
this is still in draft?

robertlong13:
> this is still in draft?

I was holding off to test the Android and mono builds just to be safe (been struggling to test it personally), and I wanted Andras to provide a translation for the new message, but I won't hold this up for it.

I'll just mark it ready.

EosBandi:
:) That was some google translate and ChatGPT...

PR #3605, opened 2025-10-26
add support for kml groundoverlay rendering

meee1:


GCSViews/FlightPlanner.cs
```diff
@@ -5193,6 +5193,198 @@ private void processKML(Element Element, Document root = null)
                     }
                 }
             }
+            else if (Element is GroundOverlay)
+            {
+                // has groundoverlay
+                var go = (GroundOverlay)Element;
+                var file = go.Icon?.Href?.ToString();
+                var latlonbox = go.GXLatLonQuad;
```
robertlong13, line 5201: We should support KMLs that don't use GXLatLonQuad, and instead use simple bounds (since that's what GoogleEarthPro uses by default). I was going to suggest a patch, but then noticed that I'd also need to account for the rotation parameter, which makes things messier, and it should also need to handle antimeridian crossing. We need to add a rather long helper function somewhere to convert LatLonBox to LatLonQuad.

At the very least, we should warn the user that we only support LatLonQuad images instead of silently failing.

robertlong13 (changes requested):

GCSViews/FlightPlanner.cs
robertlong13: [Test.zip](https://github.com/user-attachments/files/23174067/Test.zip)

meee1: the requirement for this came from atak. and is confirmed working in that context

meee1: Image

interesting options  "convert to latlonquad"

robertlong13: Correct, it's possible to convert them in Google Earth, and I tested that works fine. If that's all we want to support (which is where I lean; it's a messy pain in the ass to convert on our side), then we should pop up and tell the user to use LatLonQuad.

Issue #3509, opened 2025-03-21
Unhealthy AHRS Voice init issue


mhornsby:
There seems to be an init issue. The attached video shows the Unhealthy AHRS Voice plays until Mission Planner / Config / Planner page is opened and then it stops. I want it to be disabled so thats ok. But on startup it is not disabled.

https://github.com/user-attachments/assets/b8c5a785-b2ef-42bc-a0a2-7d5c10549a28

![Image](https://github.com/user-attachments/assets/ff949062-eaaa-46d9-9c79-5964d161fcec)

PR #3611, opened 2025-10-28
Fix issue issue #3509 Unhealthy AHRS Voice init 

mhornsby:
Fix to init issue with Speech:  "Only when armed" is set

image

### DIFF
--- a/MainV2.cs
+++ b/MainV2.cs
@@ -465,7 +465,7 @@ namespace MissionPlanner
             }
         }
 
-        public static bool speech_armed_only = truw;
+        public static bool speech_armed_only = true;
         public static bool speechEnabled()
         {
             if (speechEngine == null)
@@ -4800,4 +4800,5 @@ namespace MissionPlanner
         }
     }
 }
+
 


MainV2.cs
```diff
@@ -465,7 +465,7 @@ public static bool speechEnable
             }
         }
 
-        public static bool speech_armed_only = false;
+        public static bool speech_armed_only = truw;
```
meee1: True

mhornsby: sorry about the typo 

This is the code segment. I think the static speech_armed_only should actually be from Settings.Instance["speech_armed_only"] . But just wasn't sure and figured you would know.

The issue I was trying to resolve is the AHRS error voice message while I was doing config and testing with the rover disarmed. It was distracting .

public static bool speech_armed_only = true;
        public static bool speechEnabled()
        {
            if (speechEngine == null)
                return false;

            if (!speechEnable) {
                return false;
            }
            if (speech_armed_only) {
                return MainV2.comPort.MAV.cs.armed;
            }
            return true;
        }

mhornsby: Maybe this ?? Its not a language / enviroment I know

 public static bool speech_armed_only = false;
 public static bool speechEnabled()
 {
     if (speechEngine == null)
         return false;

     if (!speechEnable) {
         return false;
     }
     if (Settings.Instance.GetBoolean("speech_armed_only") )
         return MainV2.comPort.MAV.cs.armed;
     }
     return true;
 }

yuri-rage: @mhornsby, there's an option to simply disable speech output in the Config/Planner tab. I think that would be preferable to changing default behavior to suit an individual need.

Recommend closing this PR.

mhornsby: Hi @yuri-rage 

There's a problem with the option to disable voice while disarmed have a look at the video .

https://github.com/ArduPilot/MissionPlanner/issues/3509

 In the video speech is disabled while disarmed. But there's an init issue where that option is ignored on startup and you have to go into the config page,  and then it does whats it supposed to. That is no speech while disarmed.

https://github.com/ArduPilot/MissionPlanner/issues/3509

There is an option to completely disable speech
There is also an option to disarm speech while disarmed and that does not work.

I often work on my autopilot on the bench while its disarmed and I find the repeated unhealthy AHRS annoucement annoying.

PR #3613, opened 2025-10-31
Correct LOITER_TO_ALT plane params

robertlong13:
Revert #3477

ArduPilot source: https://github.com/ArduPilot/ardupilot/blob/d2a361afa744223354faa2cf989740458fe5b38c/libraries/AP_Mission/AP_Mission.cpp#L1163-L1168

MAVLink spec:
https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LOITER_TO_ALT

Thank you @tom-sovps for noticing this


robertlong13:
Fixed in #3614 instead

PR #3614, opened 2025-10-31
Change Crosstrack mission param display to 1=XTrack Tangent,fix error

Hwurzburg:
fix in #3477 was incorrect...thanks to @robertlong13 for catching this

### DIFF
--- a/mavcmd.xml
+++ b/mavcmd.xml
@@ -478,7 +478,7 @@
       <P1>Turns</P1>
       <P2></P2>
       <P3>Radius</P3>
-      <P4>1=Exit Tangent</P4>
+      <P4>1=XTrack Tangent</P4>
       <X>Lat</X>
       <Y>Long</Y>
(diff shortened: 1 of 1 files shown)


robertlong13:
> how about "Xtrk-tangent"

I'd go with "xtrack" over "xtrk". The letters-saved-to-clarity-lost ratio is better.

Everywhere else in mavcmd.xml uses `1=[opt]` or `[desc] (0:[opt1],1:[opt2])` for these kind of things though, and I think we should stay consistent on that.

So, all that said, I think the best option is `1=Crosstrack Tangent`. Here's a comparison of how that looks compared to the old label (and to something shorter like `1=XTrack Tangent`). At the default column widths, you can't make out the whole word without the tooltip on any of the options, but you can see the start of "crosstrack" and the start of "tangent" thanks to the line break. 

image
image
image

PR #3634, opened 2025-12-08
Fix prefetch bug - null reference in TilePrefetcherMenu

MohammadAdib:
## Summary
- Fixes null reference exception in `TilePrefetcherMenu.cs` when `Overlay` is null
- Adds null check before accessing `Overlay.Polygons` to prevent crash during tile prefetch operations

## Test plan
- [ ] Open Mission Planner and navigate to map view
- [ ] Use the tile prefetch feature
- [ ] Verify no null reference exception occurs

### DIFF
--- a/ExtLibs/GMap.NET.WindowsForms/GMap.NET.WindowsForms/TilePrefetcherMenu.cs
+++ b/ExtLibs/GMap.NET.WindowsForms/GMap.NET.WindowsForms/TilePrefetcherMenu.cs
@@ -15,19 +15,24 @@ namespace GMap.NET
 
         private RectLatLng area;
         private GMapProvider provider;
+        private bool isInitialized = false;
 
         public TilePrefetcherMenu(int min, int max, RectLatLng area, GMapProvider provider)
         {
             InitializeComponent();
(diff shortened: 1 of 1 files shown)


ExtLibs/GMap.NET.WindowsForms/GMap.NET.WindowsForms/TilePrefetcherMenu.cs
```diff
@@ -15,19 +15,24 @@ public partial class TilePrefetcherMenu : Form
 
         private RectLatLng area;
         private GMapProvider provider;
+        private bool isInitialized = false;
 
         public TilePrefetcherMenu(int min, int max, RectLatLng area, GMapProvider provider)
         {
             InitializeComponent();
-            numericUpDownMinZoom.Minimum = numericUpDownMaxZoom.Minimum = trackBarMinZoom.Minimum = trackBarMaxZoom.Minimum = min;
-            numericUpDownMinZoom.Maximum = numericUpDownMaxZoom.Maximum = trackBarMinZoom.Maximum = trackBarMaxZoom.Maximum = max;
```
robertlong13, line 23: Revert this change. I don't think moving these two lines down has anything to do with it. This PR should just be adding an isInitialized flag.

MohammadAdib: Agreed - hold my beer

PR #3641, opened 2025-12-11
Speech: fix "Only when Armed"

yuri-rage:
Returns out of `SpeakAsync()` when the user has disabled speech, fixing a bug where spoken MavLink messages meeting the severity threshold slipped through despite "Only when Armed" being selected.

There are some inconsistencies throughout the codebase in how speech is enabled/disabled. Frankly, I don't have the time to track them all down, but this fix should help homogenize behavior a bit.

The `speechEnable` property in this class was previously completely ignored and should probably be leveraged better overall.

At initial startup/connection there is a possibility that a couple of messages still slip through while disarmed. I'm not sure exactly what causes this - perhaps it takes a moment for the armed state to propagate across the application. This fix at least addresses the lion's share of nuisance messages reported.

Attached is an ArduPilot Lua script that can be used for testing.

Fixes https://github.com/ArduPilot/MissionPlanner/issues/3509

Should also close https://github.com/ArduPilot/MissionPlanner/pull/3611

[speech.zip](https://github.com/user-attachments/files/24108241/speech.zip)

### DIFF
--- a/MainV2.cs
+++ b/MainV2.cs
@@ -655,6 +655,8 @@ namespace MissionPlanner
             // load config
             LoadConfig();
 
+            speech_armed_only = Settings.Instance.GetBoolean("speech_armed_only", false);
+
             // force language to be loaded
             L10N.GetConfigLang();
 


robertlong13 (changes requested):
The addition of the `MainV2.speechEnabled()` bit is the key here. Every caller should have really been checking that function from the beginning, instead of what they've been checking, but not every caller actually has access to `MainV2`.

I think this approach is fine, I'd just drop the change on line 15. Don't worry about what I said earlier about cleaning up the callers

robertlong13 (approved):

mhornsby:
@yuri-rage I think the init is because the startup check is missing . The other speech options do this check

image

PR #3646, opened 2025-12-22
Check ADS for zone identifier before trying to load a dll plugin.

EosBandi:
Detect and report blocked plugin DLLs on Windows

On Windows 10 and 11, DLLs downloaded from the internet are marked as unsafe via a Zone Identifier (stored as an NTFS Alternate Data Stream). When a plugin DLL is blocked in this way, Windows prevents it from loading.

Previously, blocked plugin DLLs failed silently in MP, producing no error message. This led to significant user confusion and multiple support issues, especially when custom plugins were distributed to end users.

This change adds a check for the presence of a Zone Identifier = Internet on plugin DLLs. If a DLL is blocked, MP now displays a clear error message informing the user of the issue and how to resolve it.

For security reasons, the application does not attempt to remove the Zone Identifier automatically. The DLL must be manually unblocked by the user via the file’s Properties dialog.

### DIFF
--- a/Plugin/PluginLoader.cs
+++ b/Plugin/PluginLoader.cs
@@ -12,6 +12,8 @@ using DroneCAN;
 using System.Text.RegularExpressions;
 using System.Linq.Expressions;
 using System.Threading;
+using System.Runtime.InteropServices;
+using System.Text;
 
 namespace MissionPlanner.Plugin
 {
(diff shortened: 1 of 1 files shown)


meee1:
Can we do the inverse? If we know it's windows do the pinvokes, else skip. Ie positive detection. 

I'm just thinking ahead as mono is getting limited.

PR #3677, opened 2026-03-08
add Arctic/Antarctic map  projection and map sources for areas above and below 80 degrees

EosBandi:
Add map sources for polar regions with polar projections 
NASA GIBS Arctic (EPSG:3413) - Blue Marble Shaded Relief Bathymetry
NASA GIBS Antarctic (EPSG:3031) - Blue Marble Shaded Relief
Esri Polar Arctic Imagery (EPSG:5936) 15m TerraColor satellite imagery covering 50N-90N 15m/pixel
Esri Polar Arctic Ocean Base (EPSG:5936) Marine basemap with bathymetry, shaded relief, land cover, and roads
ArcGIS Arctic Bathymetry Basemap (EPSG:3995) IBCAO v4 + GEBCO_2020 bathymetry/topography with Natural Earth land cover ~66m/pixel at max zoom

Note: Grid/FaceMap will not work in these regions since UTM is not defined.

### DIFF
--- a/GCSViews/FlightData.cs
+++ b/GCSViews/FlightData.cs
@@ -55,6 +55,7 @@ namespace MissionPlanner.GCSViews
         internal GMapMarker CurrentGMapMarker;
 
         internal PointLatLng MouseDownStart;
+        internal Point MouseDownStartLocal;
 
         //The file path of the selected script
         internal string selectedscript = "";
@@ -2933,6 +2934,7 @@ namespace MissionPlanner.GCSViews
         private void gMapControl1_MouseDown(object sender, MouseEventArgs e)
         {
             MouseDownStart = gMapControl1.FromLocalToLatLng(e.X, e.Y);
+            MouseDownStartLocal = e.Location;
             Console.WriteLine("gMapControl1_MouseDown "+ MouseDownStart);
 
             if (ModifierKeys == Keys.Control)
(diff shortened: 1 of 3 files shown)


GCSViews/FlightData.cs
```diff
@@ -2971,13 +2973,42 @@ private void gMapControl1_MouseMove(object sender, MouseEventArgs e)
...
+                // incremental delta avoids direction lock near pole/singularity regions
+                int dx = e.X - MouseDownStartLocal.X;
+                int dy = e.Y - MouseDownStartLocal.Y;
 
-                double latdif = MouseDownStart.Lat - point.Lat;
-                double lngdif = MouseDownStart.Lng - point.Lng;
+                if (dx == 0 && dy == 0)
+                    return;
+
+                // optional damping near poles
```
robertlong13: I'm not sure I understand this "optional damping" part. Why slow down pan near the poles? On the new projections, the poles are perfectly stable without this, right? I commented the `absLat > 80.0` block out in FlightData and FlightPlanner and it works fine for me. Maybe I'm missing something.

Image

EosBandi: You are right, movement gain was a leftover from trying to tame the dragging around a pole. Removed it.
Also changed plus marker position calculation to use pixel position, now it is not on the rhumb line but on the actual straight line between the two markers, and projection independent.

GCSViews/FlightData.cs
```diff
@@ -2971,13 +2973,30 @@ private void gMapControl1_MouseMove(object sender, MouseEventArgs e)
...
+                double absLat = Math.Abs(gMapControl1.Position.Lat);
+                if (dx == 0) dx = Math.Sign(dx);
+                if (dy == 0) dy = Math.Sign(dy);
+
```
robertlong13: Unused variable and no-op lines (`Math.Sign(0)` is 0)
```suggestion
```

GCSViews/FlightPlanner.cs
```diff
@@ -7526,16 +7534,38 @@ private void MainMap_MouseMove(object sender, MouseEventArgs e)
...
+
+                                if (dx == 0 && dy == 0)
+                                    return;
+
+                                double absLat = Math.Abs(MainMap.Position.Lat);
+
+
+                                if (dx == 0 ) dx = Math.Sign(dx);
+                                if (dy == 0 ) dy = Math.Sign(dy);
+
```
robertlong13: Same as above
```suggestion
```

robertlong13 (changes requested):
LGTM, just a couple nits, then a rebase to fixup/squash to 2 commits, and then I'd say it's good to go.

GCSViews/FlightData.cs
EosBandi: Done. I appreciate the scrupulous review.

Issue #3673, opened 2026-02-18
Quick tab pop out window becomes a ghost window.


mperrakis2:
1. Launch Mission Planner.
2. Right click on Quick tab and select Undock. The pop out window appears.
3. Return to main Mission Planner window and click in sequence the menu strip buttons, i.e., PLAN, SETUP, CONFIG, etc. Eventually the pop out Quick window goes completely blank and becomes a ghost window.
4. Click on the DATA tool strip button to return to Flight Data screen.
5. Close the (now) Quick ghost pop out window.
6. The Quick tab goes completely blank and becomes a ghost tab as well.

ArthurPatriot:
@mperrakis2 fix already merged. @meee1 issue can be closed.

PR #3683, opened 2026-03-18
Handle Quick tab behavior when undocked in FlightData view

ArthurPatriot:
Fix for https://github.com/ArduPilot/MissionPlanner/issues/3673

When the Quick tab is undocked into a popup window, navigating away from the Flight Data screen (clicking PLAN, SETUP, CONFIG, etc.) and then returning triggers FlightData.Activate(), which calls updateDisplayView() → loadTabControlActions(). That method calls tabControlactions.TabPages.Clear() and then re-adds all tabs from the original list — including tabQuick, which is currently living in the popup form. The TabPages.Add(tabQuick) call reparents tabQuick back to tabControlactions, stealing it from the popup window and leaving it empty (the "ghost" window).
Fix (2 changes in FlightData.cs)
1.	loadTabControlActions() (line 748-749): Skip tabQuick during tab restoration when tabQuickDetached is TRUE, so it remains in the popup form.
2.	updateDisplayView() (line 772): Guard the fallback "ensure at least one tab" logic with !tabQuickDetached, preventing it from forcibly adding the detached Quick tab back.

### DIFF
--- a/GCSViews/FlightData.cs
+++ b/GCSViews/FlightData.cs
@@ -744,6 +744,10 @@ namespace MissionPlanner.GCSViews
                 bool added = false;
                 foreach (TabPage tabPage in TabListOriginal)
                 {
+                    // skip the Quick tab if it is currently undocked into a popup window
+                    if (tabQuickDetached && tabPage == tabQuick)
+                        continue;
+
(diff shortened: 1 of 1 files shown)


ArthurPatriot:
@meee1 can be merged

ArthurPatriot:
@meee1  when you have a time

PR #3685, opened 2026-03-18
Update ConfigureArducopter and FlightPlanner to accomodate new param names and units/scales

EosBandi:
**Support for ArduCopter 4.7 Parameter Changes:**

* Adds class paramchanges47 to utilities, mapping old parameter names to new ones, describes unit changes, and adds helper methods to look up mappings and warnings.
* Updated parameter setup calls in `ConfigArducopter.cs`
* Updated FlightPlanner to handle WP_RADIUS_M
* Adds a warning in the ConfigArducopter to warn users about parameter renames and unit changes when editing parameters affected by the 4.7 update. 

Note: Changes in ConfigArducopter.Designer are :
* Adding a label to display a warning
* add event handler to all NumUpDown controls' Enter event.

### DIFF
--- a/ExtLibs/Utilities/paramchanges47.cs
+++ b/ExtLibs/Utilities/paramchanges47.cs
@@ -88,11 +88,6 @@ namespace MissionPlanner.Utilities
             { "Q_LOIT_ACC_MAX_M", "Q_LOIT_ACC_MAX → Q_LOIT_ACC_MAX_M: units changed from cm/s/s to m/s/s" },
             { "Q_LOIT_BRK_ACC_M", "Q_LOIT_BRK_ACCEL → Q_LOIT_BRK_ACC_M: units changed from cm/s/s to m/s/s" },
             { "Q_LOIT_BRK_JRK_M", "Q_LOIT_BRK_JERK → Q_LOIT_BRK_JRK_M: units changed from cm/s/s/s to m/s/s/s" },
-            // Q_PILOT (Plane QuadPlane)
-            { "Q_PILOT_ACCEL_Z", "Q_PILOT_ACCEL_Z → Q_PILOT_ACCEL_Z: units changed from cm/s/s to m/s/s" },
-            { "Q_PILOT_SPD_UP", "Q_PILOT_SPEED_UP → Q_PILOT_SPD_UP: units changed from cm/s to m/s" },
-            { "Q_PILOT_SPD_DN", "Q_PILOT_SPEED_DN → Q_PILOT_SPD_DN: units changed from cm/s to m/s" },
-            { "Q_PILOT_TKO_ALT_M", "Q_PILOT_TKOFF_ALT → Q_PILOT_TKO_ALT_M: units changed from cm to m" },
             // Q_P_VELXY → Q_P_NE_VEL (Plane QuadPlane)
             { "Q_P_NE_VEL_P", "Q_P_VELXY_P → Q_P_NE_VEL_P: no scaling change, name only" },
             { "Q_P_NE_VEL_I", "Q_P_VELXY_I → Q_P_NE_VEL_I: no scaling change, name only" },
--- a/ExtLibs/Utilities/paramchanges47.cs
+++ b/ExtLibs/Utilities/paramchanges47.cs
@@ -1,0 +1,165 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace MissionPlanner.Utilities
+{
+    public class paramchanges47
+    {
+        // 4.7 parameter changes - keyed by new param name, value is "OLD → NEW: description"
+        private static readonly Dictionary<string, string> _byNewParam = new Dictionary<string, string>
+        {
+            // ARMING
+            { "ARMING_SKIPCHK", "ARMING_CHECK → ARMING_SKIPCHK: select which checks to disable" },
+            // ATC (Copter)
+            { "ATC_ANGLE_MAX", "ANGLE_MAX → ATC_ANGLE_MAX: units changed from cd to deg" },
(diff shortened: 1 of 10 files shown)


GCSViews/ConfigurationView/ConfigArducopter.cs
```diff
@@ -156,11 +156,11 @@ public void Activate()
             INS_HNTCH_OPTS.setup(0, 0, 1, 1f, new[] { "INS_HNTCH_OPTS" }, MainV2.comPort.MAV.param);
             INS_HNTCH_HMNCS.setup(0, 0, 1, 1f, new[] { "INS_HNTCH_HMNCS" }, MainV2.comPort.MAV.param);
 
-            mavlinkNumericUpDownatc_accel_r_max.setup(0, 0, 1, 0.001f, new[] {"ATC_ACCEL_R_MAX", "Q_A_ACCEL_R_MAX"},
+            mavlinkNumericUpDownatc_accel_r_max.setup(0, 0, 1, 0.001f, new[] {"ATC_ACCEL_R_MAX", "Q_A_ACCEL_R_MAX", "ATC_ACC_R_MAX", "Q_A_ACC_R_MAX"},
```
robertlong13, line 159: Don't ones like this need to handle changing the multiplier and min/max, not just adding the name to the list? Or am I missing something?

EosBandi: setup only handles one min/max/scale set for all param names listed in the array in args. I don't think it is a good idea to add scale to the entry field;  it must always show the actual param value. I tried to add the new params with the least amount of code changes. (As I see the scale arg in setup is never used, it is always zero every time) Changing the setup method would require building a full table with min/max/scale/name/increment per param name, including params back to the first supported AP version.

rmackay9: I'm perhaps not qualified to comment on MP changes but in general I also think the UI should not scale parameter values, it should just display them as-is.

robertlong13: No, I was seeing what I thought was min/max/scale/increment that was designed around old units and was being mixed with new units, which would have been bad. But my eye was off by one. I was seeing "0 to 1 with a scale of 0.001" but it's really "0 to 0 with a scale of 1 and an increment of 0.001". And now that I've looked at setup more closely, it looks like min/max are totally ignored and fetch from param meta anyway.

meee1:
for those that looked over this, are we all happy?

ExtLibs/Utilities/paramchanges47.cs
```diff
@@ -0,0 +1,165 @@
...
+            { "Q_A_ACC_P_MAX", "Q_A_ACCEL_P_MAX → Q_A_ACC_P_MAX: units changed from cm/s/s to m/s/s" },
+            { "Q_A_ACC_Y_MAX", "Q_A_ACCEL_Y_MAX → Q_A_ACC_Y_MAX: units changed from cm/s/s to m/s/s" },
+            { "Q_A_RATE_WPY_MAX", "Q_A_SLEW_YAW → Q_A_RATE_WPY_MAX: units changed from cdeg/s to deg/s" },
+            // Q_LOIT (Plane QuadPlane)
+            { "Q_LOIT_SPEED_MS", "Q_LOIT_SPEED → Q_LOIT_SPEED_MS: units changed from cm/s to m/s" },
+            { "Q_LOIT_ACC_MAX_M", "Q_LOIT_ACC_MAX → Q_LOIT_ACC_MAX_M: units changed from cm/s/s to m/s/s" },
+            { "Q_LOIT_BRK_ACC_M", "Q_LOIT_BRK_ACCEL → Q_LOIT_BRK_ACC_M: units changed from cm/s/s to m/s/s" },
+            { "Q_LOIT_BRK_JRK_M", "Q_LOIT_BRK_JERK → Q_LOIT_BRK_JRK_M: units changed from cm/s/s/s to m/s/s/s" },
+            // Q_PILOT (Plane QuadPlane)
+            { "Q_PILOT_ACCEL_Z", "Q_PILOT_ACCEL_Z → Q_PILOT_ACCEL_Z: units changed from cm/s/s to m/s/s" },
```
rmackay9, line 92: Sorry, we can remove these lines because the parameters were not changed between 4.6 and 4.7.  This mistake actually originates from my incorrect release notes, sorry! 

`
{ "Q_PILOT_ACCEL_Z", "Q_PILOT_ACCEL_Z → Q_PILOT_ACCEL_Z: units changed from cm/s/s to m/s/s" },
{ "Q_PILOT_SPD_UP", "Q_PILOT_SPEED_UP → Q_PILOT_SPD_UP: units changed from cm/s to m/s" },
{ "Q_PILOT_SPD_DN", "Q_PILOT_SPEED_DN → Q_PILOT_SPD_DN: units changed from cm/s to m/s" },
{ "Q_PILOT_TKO_ALT_M", "Q_PILOT_TKOFF_ALT → Q_PILOT_TKO_ALT_M: units changed from cm to m" },

`

rmackay9:
@EosBandi, no huge pressure of course but could you update this PR when you get a chance?  I'm just keen to clear up our [AP 4.7 issues list](https://github.com/ArduPilot/ardupilot/issues/32385)!

ExtLibs/Utilities/paramchanges47.cs
EosBandi: Done

rmackay9:
just a gentle ping on this PR.  We are getting close to releasing 4.7 (within a month) and so it would be really good to get this included

rmackay9:
great, thanks!

rmackay9:
Hmm, it seems this hasn't actually worked.

image

rmackay9:
BTW, the extended tuning screen is working in the latest beta with the exception of one field which is mentioned in this new issue: https://github.com/ArduPilot/MissionPlanner/issues/3731

Thanks!

PR #3697, opened 2026-04-02
SITL.cs: fix some swarm regressions

robertlong13:
- The `-P`/`--param` option throws an error now in ArduPilot (has been silently broken for years anyway; we didn't need them)
- `SYSID_THISMAV` is now `MAV_SYSID` in ArduPilot (setting both in identity.parm works for old/new versions)
- Driveby fix for ensuring that identity.parm is always used (wasn't in some cases)

Did a very quick test of multi-link and single-link copter swarms on AP dev.


SerhiiKorolik:
Stopped working on MacOS after this PR?

PR #3706, opened 2026-04-20
ConfigHWCompass: show calibration bad-fit reasons in mag cal dialog

christianpetri:
## Summary

Surface compass calibration failure reasons (`Mag(N) bad ...` STATUSTEXT) directly inside the onboard mag calibration dialog, so users can see *why* a calibration was rejected without leaving the page.

Pairs with the ArduPilot side that emits these diagnostics: ArduPilot/ardupilot#32757.

## What changed

In both `ConfigHWCompass` (legacy) and `ConfigHWCompass2`:

- Subscribe to `STATUSTEXT` while calibration is running (in addition to the existing `MAG_CAL_PROGRESS` and `MAG_CAL_REPORT` subscriptions).
- Filter messages with the regex `Mag\((\d+)\) bad ` and store the latest message per compass ID in a thread-safe `Dictionary<string,string>`.
- Render the latest bad-fit message per compass under the progress and report lines in `lbl_obmagresult`.
- Auto-scroll the textbox so the newest entries are visible.
- Switch line breaks to `Environment.NewLine` so newlines render correctly in the WinForms TextBox.
- Return `false` from the new STATUSTEXT branch so non-compass STATUSTEXT continues to flow to other subscribers (HUD/messages remain unaffected).
- Clear collected messages on Start; unsubscribe on Accept and Cancel.

## Why

Previously a failed compass calibration just looped silently with no indication of the cause. With ArduPilot now emitting specific `Mag(N) bad ...` reasons (bad fit, bad radius, bad orientation, etc.), Mission Planner can show these inline so the user immediately knows what to fix.

## Tests

New `MissionPlannerTests/GCSViews/ConfigHWCompassTests.cs` (MSTest):
- Regex matches each known bad-* variant and captures the compass ID.
- Regex rejects unrelated STATUSTEXT (PreArm, EKF, etc.).
- Latest-per-compass dictionary semantics validated (replacement, not accumulation).
- ASCII NUL-trim behavior validated.

Result: 7/7 passing.

## Manual verification (SITL)

- Successful calibration: progress + report lines shown, no bad-fit lines.
- Forced failure (tight `COMPASS_CAL_FIT`, extreme `SIM_MAG*_OFS_*`): `Mag(N) bad ...` lines appear in the dialog, replaced on each retry rather than accumulating.
- Multi-compass failure: one current bad line per compass ID.
- Cancel + restart: state clears cleanly.
- Normal HUD/messages stream still receives non-compass STATUSTEXT.

## Notes

- No new parameters, no MAVLink changes.
- Behavior unchanged for successful calibrations.
- AI-assisted contribution.

### DIFF
--- a/GCSViews/ConfigurationView/ConfigHWCompass.cs
+++ b/GCSViews/ConfigurationView/ConfigHWCompass.cs
@@ -6,6 +6,8 @@ using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
 using System.Drawing;
+using System.Text;
+using System.Text.RegularExpressions;
 using System.Windows.Forms;
 
 namespace MissionPlanner.GCSViews.ConfigurationView
@@ -419,6 +421,8 @@ namespace MissionPlanner.GCSViews.ConfigurationView
 
         private List<MAVLink.MAVLinkMessage> mprog = new List<MAVLink.MAVLinkMessage>();
         private List<MAVLink.MAVLinkMessage> mrep = new List<MAVLink.MAVLinkMessage>();
+        private Dictionary<string, string> badFitMessages = new Dictionary<string, string>();
+        private static readonly Regex BadFitRegex = new Regex(@"Mag\((\d+)\) bad ", RegexOptions.Compiled);
 
         private bool ReceviedPacket(MAVLink.MAVLinkMessage packet)
         {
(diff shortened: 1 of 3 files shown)


christianpetri:
Closing this in light of where ArduPilot PR #32757 ended up. The original approach here captured \`Mag(N) bad ...\` STATUSTEXT messages emitted from \`fit_acceptable()\`, but that GCS_SEND_TEXT approach was rejected during review due to flash cost (~448 bytes). The final version of #32757 emits no STATUSTEXT — it only sets new enum values (BAD_OFFSETS=8, BAD_DIAG=9, BAD_FITNESS=10) in \`MAG_CAL_REPORT.cal_status\`. That makes the STATUSTEXT subscription, regex, and dictionary added here dead code that will never fire. The right fix for issue #3709 is a much smaller change: add the three new values to the \`MAG_CAL_STATUS\` enum in \`ExtLibs/Mavlink/Mavlink.cs\`. The existing \`(MAVLink.MAG_CAL_STATUS)obj.cal_status\` cast in both compass forms will then display named labels automatically, with no extra subscription or logic needed. Will open a replacement PR once the upstream mavlink/mavlink XML is merged.

PR #3713, opened 2026-04-27
Fix time calculation logic by replacing `.Milliseconds` with `.TotalMilliseconds`

ArthurPatriot:
This PR fixes a logical bug in time interval calculations where the fractional .Milliseconds property was used instead of the cumulative .TotalMilliseconds.

In .NET, TimeSpan.Milliseconds only returns the millisecond component of the interval (0-999). If an interval exceeds 1 second, the comparison logic would fail or behave inconsistently (e.g., an interval of 1100ms would be treated as 100ms).


ArthurPatriot:
@meee1 when you have a time

ArthurPatriot:
@meee1 when you have a time

PR #3714, opened 2026-04-27
Fix event handler leaks in FlightData, FlightPlanner, and adsb

ArthurPatriot:
Event handlers were either never unsubscribed on disposal or accumulated unboundedly on a static event, preventing garbage collection of the affected objects.


ArthurPatriot:
@meee1 when you have a time

PR #3715, opened 2026-04-27
 Fix HttpClient misuse: reuse static instance, remove duplicate headers, add missing error handling

ArthurPatriot:
HttpClient was being instantiated on every request across Download, Tracking, and adsb. This is a well-known anti-pattern that exhausts the system's socket pool under load (each disposed instance holds a socket in TIME_WAIT for up to 4 minutes).


ArthurPatriot:
@meee1 when you have a time

PR #3718, opened 2026-05-04
MavlinkInterface: Fix guided mode on copter to use alt reference frame 

EosBandi:
Copter always used relative frame instead of the ref.frame set in the FlyToAlt window.

### DIFF
--- a/ExtLibs/ArduPilot/Mavlink/MAVLinkInterface.cs
+++ b/ExtLibs/ArduPilot/Mavlink/MAVLinkInterface.cs
@@ -4448,7 +4448,7 @@ Mission Planner waits for 2 valid heartbeat packets before connecting
                 else
                 {
                     setPositionTargetGlobalInt((byte) sysid, (byte) compid,
-                        true, false, false, false, MAVLink.MAV_FRAME.GLOBAL_RELATIVE_ALT_INT,
+                        true, false, false, false, (MAV_FRAME)gotohere.frame,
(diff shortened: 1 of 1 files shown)


robertlong13 (approved):
Damn, I can't believe I missed that

MohammadAdib:
Heads-up that this change has a subtle ripple effect for setGuidedModeWP callers that don't explicitly set
  Locationwp.frame. Since Locationwp.frame defaults to 0 (= MAV_FRAME.GLOBAL, i.e. AMSL), the previously hardcoded
  GLOBAL_RELATIVE_ALT_INT was silently normalizing those calls. With this fix in place, the same calls now go out as
   AMSL on Copter, which can mean a commanded descent below ground if the caller was passing a relative altitude.

  Quick grep of the affected sites in the repo:

  - Controls/FollowMe.cs:177 — Follow Me builds Locationwp with only id/alt/lat/lng
  - Utilities/httpserver.cs:740 and :776 — the /guided HTTP endpoints
  - Swarm/FollowPath.cs:50
  - ExtLibs/ExtGuided/ExtGuidedPlugin.cs:61
  - Plugins/example14-mass.cs:129

  Each of these would need frame = (byte)MAV_FRAME.GLOBAL_RELATIVE_ALT (or whatever the intended frame actually is)
  added to preserve pre-fix behavior. The FlyToHere path in FlightData.cs is fine since it threads GuidedMode.frame
  through.

@robertlong13 @EosBandi @meee1

Issue #3696, opened 2026-03-31
Bitmask window incorrectly calculating values past 2^23 causing incorrect parameter setting


AlfieLockrey:
When using the pop out bitmask in the Full Parameter List, the bitmask value is incorrectly calculated when selecting servo 25 and above:

<img width="588" height="216" alt="Image" src="https://github.com/user-attachments/assets/f88dca1c-2fb2-49d0-8284-883e99f1abb9" />

<img width="729" height="383" alt="Image" src="https://github.com/user-attachments/assets/c6bb2a10-4591-4b1d-bd12-7b59bfb582a9" />

If you close and reopen the bitmask window, it will correctly read the incorrectly written value:

<img width="741" height="388" alt="Image" src="https://github.com/user-attachments/assets/88b42737-8a20-4099-9755-3df07d3c9c3b" />

A work around exists by manually entering the numerical bitmask value calculated seperately, however if the bitmask window is interacted with, it will incorrectly update the bitmask window before closing.

I have had this present issues when operating a UA with the motor channels mapped to servos 20-27. It resulted in a servo channel being incorrectly unselected and not being passed through to CAN (CAN_D1_UC_SRV_BM).

This issue also exists for the servo bitmask option in the same way so I am presuming it is an error with bitmasks past a certain size.

robertlong13:
There's a fundamental limitation with ArduPilot actually, all parameter values are sent as float32 over MAVLink (which has a 23 bit mantissa). After you set bits higher than 23, you'll start to get other random low bits set, essentially as "rounding errors".

Now, the issue you're seeing here is that Planner is doing the same float32 thing with that dialog, but I'm pretty sure that if you write that value, then read it back, I'm guessing you'll get the same problem. If so, I'm not sure I can think of a good way around this other than custom building with defaults.parm

AlfieLockrey:
@robertlong13 I have done some further testing and have no issues sending the higher value bitmasks over mavlink when entering the bitmask value manually. I can read and write the param without issue to the vehicle. The issue only arises when a checkbox on the bitmask window has been interacted with and the bitmask recalculated from the window. I would assume that MAVLINK and MPs param sending code is handling the bitmask correctly as it functions when entered numerically by hand.

robertlong13:
Okay, thanks for checking that. I'll look into this

IamPete1:
ftp param read/write avoids the issues with float transport.

PR #3723, opened 2026-05-14
MavlinkCheckBoxBitMask: fix bitmask value formatting

tatsuy:
Fiexed #3696.
This is primarily a Mission Planner Set Bitmask UI issue.

Adding a small bit to a bitmask value greater than 10,000,000 makes this phenomenon easier to reproduce.

The bitmask popup previously built the selected value through a float and returned it via Value.ToString(). For larger bitmask values the displayed parameter value could be rounded before being written back to the parameter cell.

[Before]
SIM_CAN_SRV_MSK
	Servo 25 -> 16777220
	Servo 3 + Servo 25 -> 16777220

[After - This PR]
SIM_CAN_SRV_MSK	16777216
	Servo 25 -> 16777216
	Servo 3 + Servo 25 -> 16777220

### DIFF
--- a/Controls/MavlinkCheckBoxBitMask.cs
+++ b/Controls/MavlinkCheckBoxBitMask.cs
@@ -1,6 +1,7 @@
 ﻿using MissionPlanner.Utilities;
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Windows.Forms;
 
 namespace MissionPlanner.Controls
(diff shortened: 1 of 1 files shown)


tatsuy:
@meee1 Could you look at this if you have time?
We can reproduce this easily using SITL and real FC.
image

image

Problems occur with certain parameters, such as FLTMODE_GCSBLOCK and CAN_D1_UC_SRV_BM.

PR #3725, opened 2026-05-21
FlightPlanner: add DO_RETURN_PATH_START to mission editor

tridge:
DO_RETURN_PATH_START (MAV_CMD 188) was missing from mavcmd.xml, so the mission editor showed it as UNKNOWN and failed to load missions containing it. Add it to the AC2 and APM sections (matching DO_LAND_START) in both the main and Xamarin command lists, and render it on the map as a located waypoint when it carries a lat/lng.


robertlong13:
I have this in  https://github.com/ArduPilot/MissionPlanner/pull/3668

Along with actually rendering the return path as a separate color.

tridge:
> I have this in #3668

ok, didn't know that, sorry. 
I hit this today with a user. Couldn't use DO_RETURN_PATH_START at all. Maybe the minimal fix should go in and try to get the larger fix in later?
@meee1 ?

PR #3744, opened 2026-06-23
Custom actions via plugin

Ryszard-Trojnacki:
This PR allows to insert custom actions via plugin to Action tab.

image

Usage (inside plugin):
```c#
        public override bool Loaded()
        {
            Host.MainForm.FlightData.RegisterCustomAction("Custom_action_name", (a) =>
                {
                    // Custom action code
                }
            );
        }
```

Change by this PR in `FlightData.cs`:
1. Added `private BindingList<string> ActionList=new BindingList<string>(Enum.GetNames(typeof(actions)).ToList());` to have mutable actions.
2. Added `private Dictionary<string, Action<string>> CustomActions = new Dictionary<string, Action<string>>();` mapping from name to handler.
3. Changed `void BUTactiondo_Click(object sender, EventArgs e)` to check if custom action was executed and if so, call registered handler.
4. Function to register custom action `void RegisterCustomAction(string action, Action<string> handler, string after=null, string before=null)` and unregister `bool UnregisterCustomAction(string action)` that just modifies `ActionList` and `CustomActions`.

Changed code doesn't break current behaviour. It just adds a way to add something to this part of application.

### DIFF
--- a/GCSViews/FlightData.cs
+++ b/GCSViews/FlightData.cs
@@ -15,6 +15,7 @@ using MissionPlanner.Utilities;
 using MissionPlanner.Warnings;
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Diagnostics;
 using System.Drawing;
 using System.Drawing.Drawing2D;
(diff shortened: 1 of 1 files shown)


Autonomost:
@meee1 the fix for this should get merged in. Everyone got bit by this since this commit. unless I missed the fix somehow?

PR #3760, opened 2026-08-20
httpserver: restrict /websocket/raw to localhost

### LINKED ISSUE #3757: Is the MAVLink WebSocket on TCP/56781 intended to be remotely writable?
I found that Mission Planner starts a WebSocket server on TCP port 56781 using non-loopback network interfaces. One of the exposed endpoints is "/websocket/raw". The endpoint does not require authentication, and data received is written directly to Mission Planner's active vehicle connection.

I was able to connect to this endpoint from another host on the network and send a MAVLink command to a connected ArduCopter SITL instance. For example, I was able to change the disarmed vehicle from STABILIZE to ALT_HOLD.

My main question is whether remote write access to "/websocket/raw" is intended. Looking at the code, it seems to hint towards an intention for read only and localhost only. The surrounding Mission Planner/ArduPilot usage I found appears to treat port 56781 as a localhost integration interface. However, I do not have clear proof for that.

Greninja44:
Fixes https://github.com/ArduPilot/MissionPlanner/issues/3757
The /websocket/raw WebSocket endpoint on TCP/56781 accepts unauthenticated connections from any network interface. Remote hosts can inject arbitrary MAVLink commands into the active vehicle link — for example, changing the flight mode of a disarmed drone.

This adds an IPAddress.IsLoopback() check on the TCP remote endpoint before completing the WebSocket handshake, consistent with the existing protections on /guided? and POST /guide (added in ec2b6fcda and 980d2b8ab).

Applied to both the desktop server (Utilities/httpserver.cs) and the Xamarin copy (ExtLibs/Xamarin/Xamarin/Linked/httpserver.cs).

All built-in clients (WASM, HUD, Blazor) connect to localhost and are unaffected.


fallenmi (approved):
I verified that the new guard runs before the WebSocket `101` response and before the raw bridge is installed in both the desktop and Xamarin copies. A socket-level oracle using the same `TcpClient.Client.RemoteEndPoint` / `IPAddress.IsLoopback` path sees `127.0.0.1` as loopback and a connection through the host's non-loopback interface as non-loopback, so the latter takes the fail-closed 403 path while the built-in local clients remain allowed. A missing or non-IP endpoint is also rejected.

The net PR diff is limited to these two matching guards. Release, Debug, macOS, and iOS builds are green. The two Android failures are baseline infrastructure failures: exact current base `67a3c4f` and this head both fail because the VS18 runner lacks `Xamarin.Android.CSharp.targets`, with the same MSB4226 error.

This closes the unauthenticated remote-command path without changing the local WebSocket protocol. Looks good to me.

Disclosure: I used OpenAI Codex to inspect the exact head/base, verify the socket address classification, and compare the CI failures; I verified the result.

PR #3780, opened 2026-09-17
Plan: add an ALL mission type to edit mission, fence and rally items …

Autonomost:
…together

The mission type dropdown on the Plan screen gains an ALL entry. In that mode the single command grid holds mission, fence and rally items at once; a row's type follows its command (RALLY_POINT, FENCE_* or anything else), so no extra columns are needed. The map draws all three sets in one overlay with type-aware midline handles, and the fence polygon tools work in this mode.

Read and Write open a small picker (Mission / Fence / Rally, all ticked by default) so one click transfers everything or any subset. Reading a subset keeps the grid rows of the types that were not fetched. Uploads go in mission, fence, rally order; a type the vehicle rejects (for example fence items on old firmware) is reported without aborting the rest.

The MAVFTP download path now updates the cache matching the item type instead of always writing the mission cache.


Autonomost (commented):
Made the changes as requested.

PR #3781, opened 2026-09-17
Plan/Data/SITL: enter and show home, POIs and SITL start in GEO, UTM …

Autonomost:
…or MGRS

Home Location on the Plan screen follows the coordinate system chosen in the mouse position readout: in UTM or MGRS the Lat/Long boxes are replaced by one box showing the same point, and typing a value there converts back to lat/long, which the rest of the planner keeps using.

A shared CoordsInputBox dialog (frame picker, hint line, coordinates, optional altitude) replaces the ad hoc "lat;long;alt" prompt for Add POI > Coords on the Data screen. POI labels are built when drawn, in the frame selected on that screen, POI altitude is now persisted, and a right-click Add POI fills in terrain height at the point. The altitude lookup for a typed POI now uses the POI position rather than the mouse position.

The Simulation screen gains an "Enter Home..." button so the SITL start location can be typed in any of the three frames instead of dragged, and the map group title shows the current start position.


Autonomost (commented):
Made the changes as requested.

